PR #195, opened 2020-09-03
fix: Migrate config dir to XDG_CONFIG_HOME

### LINKED ISSUE #186: Support XDG Base Directory spec for configuration files
**Is your feature request related to a problem? Please describe.**
There is a [spec](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) for where different type of files should be. The main thing to this project is the configuration files. This allows for decluttering the home directory and having all the config files in one place.

**Describe the solution you'd like**
Glab should try to read configuration files from directory specified in `$XDG_CONFIG_HOME` or default to `$HOME/.config` if this environment variable is empty as per spec.

If there is no configuration there it should still try to read form `$HOME/.glab-cli` like it does now to not break backwards compatibility.

andhe:
**Note: bugfix included**
While doing this a bug in the original code was spotted where the
wrong error variable would be printed on a fatal error:

```
-                               log.Fatalln(err)
+                               log.Fatalln(errDir)
```

**Related Issue**

Fixes #186

**Motivation and Context**

Using the legacy scheme of every application putting files in the users home directory where ever they preferred caused alot of clutter and no consistency. For example erasing all cache files to free up disk space and/or backing up (only) config files would require intimate knowledge of every applications way to save their files. XDG Base Dir Specification was created to solve these issues. With the changes included here we migrate the old files over to use the XDG Base Dir spec location and use it.

**How Has This Been Tested?**

Only this trivial test was performed (using the new code, creating a config file, manually moving the file back to the old location, then triggering a migration):

```
make build
./bin/glab --global config
mkdir -p ~/.glab-cli/config
mv ~/.config/glab-cli/.env ~/.glab-cli/config/
rmdir ~/.config/glab-cli

./bin/glab --help 2>&1 | head -n 1

ls -lRa ~/.glab-cli/ ~/.config/glab-cli/
```

Note: no alias file testing has been done. Help with further testing welcome!

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -8,6 +8,7 @@ import (
 	"log"
 	"os"
 	"os/user"
+	"path/filepath"
 	"strings"
 
 	"github.com/profclems/glab/internal/manip"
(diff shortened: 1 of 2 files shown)


profclems (approved):
Thanks @andhe! LGTM!

PR #200, opened 2020-09-04
fix(config): Make SetEnv invalidate cache used by GetEnv

andhe:
The golden rule about caches is that if there's no way
to invalidate them then they're broken...

And indeed if config.SetEnv("foo", ...) is called followed
by config.GetEnv("foo") then you'll get the old value (when
both read/write from the same file). And example of this
is on clean start where initConfig sets up the default
values for missing/empty env vars.

A testcase for this:

```
rm -rf ~/.config/glab-cli/ .glab-cli/config/.env

./bin/glab config --global
```

Before this patch it would say:

```
? Enter default Gitlab Host (Current Value: ):
```

After patch it correctly lists the default/current value
as set by initConfig():

```
? Enter default Gitlab Host (Current Value: https://gitlab.com):
```

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -128,6 +128,9 @@ func SetEnv(key, value string) {
 	if UseGlobalConfig {
 		cFile = globalConfigFile
 	}
+
+	defer InvalidateEnvCacheForFile(cFile)
+
 	data, _ := ioutil.ReadFile(cFile)
 
 	file := string(data)
--- a/internal/config/file.go
+++ b/internal/config/file.go
@@ -56,6 +56,11 @@ func readConfig(filePath string) map[string]string {
 	return config
 }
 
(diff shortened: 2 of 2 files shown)


andhe:
(On second thought I pushed an updated commit. It felt safer to defer the invalidation to when the new env file had actually been written out for the theoretical case of SetEnv/GetEnv running in parallel. Invalidating the cache before writing out the new content means there's a race. Not a practical problem for glab since it doesn't do things in parallel, but anyway....)

profclems:
> (On second thought I pushed an updated commit. It felt safer to defer the invalidation to when the new env file had actually been written out for the theoretical case of SetEnv/GetEnv running in parallel. Invalidating the cache before writing out the new content means there's a race. Not a practical problem for glab since it doesn't do things in parallel, but anyway....)

I thought so too... The tests are parallelly executed

profclems (approved):
Thanks @andhe **LGTM**! You spotted this issue much earlier. I never thought of this. It would have been an issue

PR #201, opened 2020-09-04
fix(config): SetEnv appends to gitignore every time

andhe:
**Description**
This change fixes so that my .gitignore doesn't grow an additional ".glab-cli" line every time I run the command (which thus far has been only to test `./bin/glab config --global`).

**How Has This Been Tested?**
I've run `rm -rf ~/.config/glab-cli ; ./bin/glab config --global` under the following circumstances and observed the behaviour:
* Default `.gitignore` already containing `.glab-cli` -> without fix: .glab-cli appended to .gitignore twice(!) -> with fix: no change of .gitignore file
* Both(!) occurances of `.glab-cli` removed from `.gitignore` -> without fix: .glab-cli appended to .gitignore twice(!) -> with fix: .glab-cli appended to .gitignore
* With `.gitignore` file removed -> without fix: .gitignore created and .glab-cli appended to .gitignore twice(!) -> with fix: .gitignore created and .glab-cli appended to .gitignore

Note: testing was done *before* the second commit, obviously using --global flag for test would not be useful.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -156,7 +156,7 @@ func SetEnv(key, value string) {
 	w := bufio.NewWriter(f)
 	_, _ = w.WriteString(strings.Trim(newData, "\n"))
 	_ = w.Flush()
-	if !CheckFileHasLine(".gitignore", configFileFileParentDir) {
+	if !UseGlobalConfig && !CheckFileHasLine(".gitignore", configFileFileParentDir) {
 		ReadAndAppend(".gitignore", configFileFileParentDir)
 	}
 }
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -156,7 +156,7 @@ func SetEnv(key, value string) {
 	w := bufio.NewWriter(f)
 	_, _ = w.WriteString(strings.Trim(newData, "\n"))
 	_ = w.Flush()
-	if GetKeyValueInFile(".gitignore", configFileFileParentDir) == "NOTFOUND" {
+	if !CheckFileHasLine(".gitignore", configFileFileParentDir) {
 		ReadAndAppend(".gitignore", configFileFileParentDir)
 	}
 }
--- a/internal/config/file.go
+++ b/internal/config/file.go
@@ -1,6 +1,7 @@
 package config
 
 import (
+	"bufio"
 	"io/ioutil"
 	"log"
 	"os"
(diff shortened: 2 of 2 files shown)


andhe:
Additional idea: Would likely be a good idea to add an additional fix on top of this to check if current working directory (from where glab command is run) is actually a git repo (i.e. .git exists in current (or parent) directory) at all before creating a .gitignore file to avoid creating clutter in filesystem.

Testcase:
```
$ export T=$(mktemp -d)
$ echo $T
/tmp/tmp.AG9MnbzXtK
$ cd $T
$ rm -rf ~/.config/glab-cli ; ~/go/src/github.com/profclems/glab/bin/glab config --global
? Enter default Gitlab Host (Current Value: ):  
2020/09/05 01:48:07 interrupt
$ ls -la
total 4
drwx------  2 ah   ah     60 sep  5 01:48 .
drwxrwxrwt 54 root root 2560 sep  5 01:47 ..
-rw-r--r--  1 ah   ah     20 sep  5 01:48 .gitignore
$ cat .gitignore 

.glab-cli
.glab-cli
```

Possibly easier fix, but slight change in behaviour would be to only add to .gitignore if such file already exists.

andhe:
Another thing I spotted in config.SetEnv is that it opens a file descriptor via os.Create but never closes it and thus relies on the GC to spot all references has gone out of scope and eventually close the fd. Would probably be a good idea to explicitly close the file descriptor (right after w.Flush() is called) to help with making sure the changes actually goes out to the filesystem (and not only flushing the bufio Writers cache).

On top of that, writing changes to a temporary file first (including handling errors!) and then renaming/moving the file over the original to atomically replace it would be an improvement. For example the current code could cause data loss in out-of-diskspace situations.... Apparently there's a package to help with this: https://github.com/google/renameio

profclems:
> Additional idea: Would likely be a good idea to add an additional fix on top of this to check if current working directory (from where glab command is run) is actually a git repo (i.e. .git exists in current (or parent) directory) at all before creating a .gitignore file to avoid creating clutter in filesystem.

I agree with this. 

> Possibly easier fix, but slight change in behaviour would be to only add to .gitignore if such file already exists.

I thought of moving the local config file into the `.git` folder but that will be best if considered for v2.0. 
We can check if the .git folder exists before creating/appending to the .gitignore. The local config directory also shouldn't be created at all if the directory isn't a git repo. 
Another way to do this is to check for the current branch with `git.CurrentBranch()` found in
https://github.com/profclems/glab/blob/b07ce916173d3f59872f5ac7ba457a3658dc43d7/internal/git/git.go#L157

profclems:
> On top of that, writing changes to a temporary file first (including handling errors!) and then renaming/moving the file over the original to atomically replace it would be an improvement. For example the current code could cause data loss in out-of-diskspace situations.... Apparently there's a package to help with this: https://github.com/google/renameio

The package looks promising. We should use that then.
It will be best to do that in a separate PR

profclems (approved):

PR #202, opened 2020-09-05
fix: File content modification made a bit more safe and sane

andhe:
**Description**
Atomically replace file that was previously truncated and then modified content written out again (without any error handling). New code also does some basic error handling. Replacing the old code with using renameio.WriteFile instead also got rid of the missing/implicit file Close(). See commit message for futher details!

While at it also fix up the ReadAndAppend method usage and implementation that is used to extent .gitignore so that it behaves better with regard to newline handling.

**Testing**
Some trivial manual testing has been done to try to verify the code works as expected, but could certainly be better and more extensively.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -10,6 +10,7 @@ require (
 	github.com/fatih/color v1.9.0 // indirect
 	github.com/fsnotify/fsnotify v1.4.9 // indirect
 	github.com/gdamore/tcell v1.4.0
+	github.com/google/renameio v0.1.0
 	github.com/google/shlex v0.0.0-20191202100458-e7afc7fbc510
 	github.com/gookit/color v1.2.9
 	github.com/gosuri/uilive v0.0.4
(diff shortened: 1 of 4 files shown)


profclems (approved):
Nice one... Thanks!

profclems:
@andhe, I would like to packaqge `glab` for debian so it can be installed with `sudo apt install glab`. Can you help out with that?
I may as well open it as an issue

andhe:
sure thing. I'll start looking into if all dependencies are already packaged when i find the time later. We should be able to get a work needed estimate with `dh-make-golang estimate github.com/profclems/glab`.

profclems:
I guess there's one missing dependency
```
$ dh-make-golang estimate github.com/profclems/glab
go get: 185.08 MiBpackage github.com/russross/blackfriday/v2: cannot find package "github.com/russross/blackfriday/v2" in any of:
        /usr/local/go/src/github.com/russross/blackfriday/v2 (from $GOROOT)
        /tmp/dh-make-golang243160843/src/github.com/russross/blackfriday/v2 (from $GOPATH)
2020/09/05 16:42:42 exit status 1
```

PR #211, opened 2020-09-07
fix(config): use ioutil.WriteFile instead of renameio

### LINKED ISSUE #210: Failed to update config file on windows
Running `glab config -t ...` failes with `Failed to update config file: chmod C:\Users\...\AppData\Local\Temp\..env337738213: not supported by windows`

**Description**
https://github.com/profclems/glab/blob/3ff88b6ea5bb8e2aba83d6e29f9cefac533a0979/internal/config/config.go#L162
renameio.WriteFile uses [File.Chmod](https://golang.org/pkg/os/#File.Chmod) which is not supported on windows.

**Expected Behavior**
`glab config` command can write the config file

**Actual Behavior**
`glab config` failes when writing the config file

**Possible Fix**
Remove renameio and add some error handling to the open-write go code.
I think that should be enough to guarantee (99.99... %) not to destroy the config file.
@andhe please correct me if i'm wrong.
Alternative, fix https://github.com/google/renameio/issues/17

**Steps to Reproduce**
1. run `glab config -t 123`

**Your Environment**
* Version used: Branch:trunk
* Operating System and version: Windows 10

andhe:
As reported apparently renameio does not work on windows and there's
no obvious way to even make an atomic file overwrite on windows at all,
thus renameio is looking at dropping Windows support.

This change simply switches to ioutil.WriteFile which is not atomic
but the code should still be nicer than the previous one we replaced.

While at it also cut down on the readability part of the files
permissions as the config file could contain sensitive information
(like the token).

Fixes #210


profclems (commented):
I don't know if this is a right decision to make the atm.

> I'm thinking the need of importing/linking runtime (and renameio on windows) could be avoided by creating a WriteFile helper method in a file which has a !+build windows and another writefile_windows.go which implements WrifeFile using ioutil.WriteFile.

This looks like a fairly optimal solution for me.

andhe:
Ok, published #212 as a replacement for this.

PR #212, opened 2020-09-07
fix(config): Use ioutil.WriteFile on Windows

### LINKED ISSUE #210: Failed to update config file on windows

andhe:
As reported renameio doesn't work on Windows and apparently there's
no way to do atomic file replace on Windows.

We thus create a helper function in a separate file that uses renameio
but say it should not be built on windows, then hook in a windows-
specific implementation of that function which uses ioutil.WriteFile
which is not safe but gets the job done.

While at it also tighten up the permissions on the config file as
it might contain sensitive information (like the token).

Fixes #210

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -12,7 +12,6 @@ import (
 
 	"github.com/profclems/glab/internal/manip"
 
-	"github.com/google/renameio"
 	"github.com/logrusorgru/aurora"
 	"github.com/spf13/cobra"
 	"github.com/tcnksm/go-gitconfig"
@@ -159,7 +158,7 @@ func SetEnv(key, value string) {
 		newData += newConfig
 	}
 	_ = os.MkdirAll(filepath.Join(cFile, ".."), 0755)
-	if err = renameio.WriteFile(cFile, []byte(newData), 0666); err != nil {
+	if err = WriteFile(cFile, []byte(newData), 0600); err != nil {
 		log.Println("Failed to update config file:", err)
 		return
 	}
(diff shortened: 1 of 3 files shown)


andhe:
Note that this has not actually been tested on windows (except GOOS=windows cross-build).

andhe:
I also have no idea how to make renameio in go.sum not apply on windows. I guess that means people building on Windows will still have to download renameio even if it will not be used in the build?

profclems (commented):
Thanks @andhe!
I will build and test on windows and merge if it passes.

profclems (approved):

PR #244, opened 2020-10-02
fix: glab pipeline list always says "No pipelines..."

andhe:
This commit fixes up previous commit 6c4fe1c2e82fd36f086b3
"Isolate commands into separate sub-packages (#229)"

**Description**
Before the `glab pipeline list` command would always result in:
`No Pipelines available on project/path`

**How Has This Been Tested?**
Running on a project path which I knew had pipelines. With the fix they show up.


profclems (approved):
Nice catch @andhe

PR #245, opened 2020-10-03
feat: support "Run pipeline"

andhe:
**Description**

The API calls it "create pipeline", but the gitlab
web interface button to do this says "Run pipeline".

I named the command run as that's likely to be more familiar
to users.

With this you can now support starting a pipeline via
"glab pipeline run -b mybranch".

Note that setting Ref when calling the CreatePipeline API
seems required or we'll end up with an "Internal Server Error".
We thus do a best-effort attempt at figuring out the default
when the branch argument is not specified.

**Motivation and Context**
I have pipelines that I need to regularly trigger manually. Rather than using the "Run pipeline" button in the web interface it would be convenient to be able to do this via the cli.

**How Has This Been Tested?**
I triggered a pipeline run on one of my repos and looked at the web interface to see it run.

Please help test this on a repo where the default branch is not named "master" (and "master" doesn't exist at all).

**Types of changes**

### DIFF
--- a/commands/pipeline/run/pipeline_run.go
+++ b/commands/pipeline/run/pipeline_run.go
@@ -38,6 +38,7 @@ func NewCmdRun(f *cmdutils.Factory) *cobra.Command {
 	var pipelineRunCmd = &cobra.Command{
 		Use:   "run [flags]",
 		Short: `Create a new pipeline run`,
+		Aliases: []string{"create"},
 		Example: heredoc.Doc(`
 	$ glab pipeline run
 	$ glab pipeline run -b trunk
--- a/commands/pipeline/run/pipeline_run.go
+++ b/commands/pipeline/run/pipeline_run.go
@@ -1,0 +1,97 @@
+package run
+
+import (
+	"fmt"
+
+	"github.com/profclems/glab/commands/cmdutils"
+	"github.com/profclems/glab/internal/git"
+	"github.com/profclems/glab/internal/utils"
+	"github.com/profclems/glab/pkg/api"
+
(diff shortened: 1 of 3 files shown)


commands/pipeline/run/pipeline_run.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+
+	branch, _ := git.GetDefaultBranch(repoRemote.Name)
+
+	return branch
+}
+
+func NewCmdRun(f *cmdutils.Factory) *cobra.Command {
+	var pipelineRunCmd = &cobra.Command{
+		Use:   "run [flags]",
```
profclems, line 39: Since the API uses 'create', we should add `create` as an alias to the run command with:
```go
.....,
Aliases: []string{"create"},
```

profclems:
@andhe I sent you a collaborator invitation to help maintain this project. Let me know if you're up for it

andhe:
> @andhe I sent you a collaborator invitation to help maintain this project. Let me know if you're up for it

Thanks. I'm not sure exactly what that means. I'm quite a noob on both GitHub and Go really. I personally find it good if you can review my PRs, but I understand if you might want help. Not sure where I stand on taking on any (more) responsibilities right now.

profclems:
> > @andhe I sent you a collaborator invitation to help maintain this project. Let me know if you're up for it
> 
> Thanks. I'm not sure exactly what that means. I'm quite a noob on both GitHub and Go really. I personally find it good if you can review my PRs, but I understand if you might want help. Not sure where I stand on taking on any (more) responsibilities right now.

profclems (approved):

PR #257, opened 2020-10-08
Feature:  mr list by asssignee

### LINKED ISSUE #167: Ability to list merge requests assigned to a user
**Is your feature request related to a problem? Please describe.**
There should be additional flags on the `glab mr list` command to list merge request assigned to current user or a specific user

**Describe the solution you'd like**
Add --mine and --assignee flags to `glab mr list`

To get list of MRs for the current user:
```sh
glab mr list --mine
```
To get the list of MRs for a user with username
```sh
glab mr list --assignee=<username>
```
Should allow multiple assignees listing
```sh
glab mr list --assignee=user1,user2
```

zemzale:
**Description**

Added `mr list` option `--assingee` to get MRs for specific assignees.

To do this added 3 API functions :
- `ListMRsWithAssignees` - allows us to get MR by their assignee IDs
- `UserByName` - allows to get a single user by it's name
- `UsersByNames` - allows to get multiple users by names

Now if user passes `--assigne=use`r or `--assigne=user`,user2` flag to the command, it's gonna show up MRs that are assigned to the users.  
It will fail with error message that we can't find any of the usernames that were passed to the command.

There is one issue that is not related to filtering of the users but displaying the MRs. There is no way for the end user to know which MR is assigned to whom. That probably deserves it's own issue since there are multiple ways of how that could be done e.g. new column or multiple table for each assignee. 
I tried to just add new column but every row overflowed and since I got some time constrains I didn't investigate further for now.

**Related Issue**

#167  
**Motivation and Context**

Allow better filtering for MR listing
**How Has This Been Tested?**

Tested on a private GitLab repo, with single and multiple assignees.

The only issue I can see with this is that if you add multiple assignees and provide the `--all` flag it will show 20 MRs for each assignee if they have that many to them (which can be pretty normal for a bit bigger projects) 

**Types of changes**

**Checklist:**

- [ ] All new and existing tests passed.

### DIFF
--- a/pkg/api/merge_request.go
+++ b/pkg/api/merge_request.go
@@ -52,6 +52,22 @@ var ListMRs = func(client *gitlab.Client, projectID interface{}, opts *gitlab.Li
 	return mrs, nil
 }
 
+var ListMRsWithAssignees = func(client *gitlab.Client, projectID interface{}, opts *gitlab.ListProjectMergeRequestsOptions, assigneeIds []int) ([]*gitlab.MergeRequest, error) {
+	if client == nil {
+		client = apiClient
+	}
+	mrs := make([]*gitlab.MergeRequest, 0)
+	for _, id := range assigneeIds {
+		opts.AssigneeID = gitlab.Int(id)
+		assingeMrs, _, err := client.MergeRequests.ListProjectMergeRequests(projectID, opts)
+		if err != nil {
+			return nil, err
+		}
+		mrs = append(mrs, assingeMrs...)
+	}
+	return mrs, nil
+}
+
 var UpdateMR = func(client *gitlab.Client, projectID interface{}, mrID int, opts *gitlab.UpdateMergeRequestOptions) (*gitlab.MergeRequest, error) {
 	if client == nil {
 		client = apiClient
--- a/pkg/api/merge_request.go
+++ b/pkg/api/merge_request.go
@@ -59,7 +59,7 @@ var ListMRsWithAssignees = func(client *gitlab.Client, projectID interface{}, op
 	mrs := make([]*gitlab.MergeRequest, 0)
 	for _, id := range assigneeIds {
 		opts.AssigneeID = gitlab.Int(id)
-		assingeMrs, _, err := client.MergeRequests.ListProjectMergeRequests(projectID, opts)
+		assingeMrs, err := ListMRs(client, projectID, opts)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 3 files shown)


pkg/api/user.go
```diff
@@ -14,3 +17,17 @@ var CurrentUser = func(client *gitlab.Client) (*gitlab.User, error) {
...
+
+var UserByName = func(client *gitlab.Client, name string) (*gitlab.User, error) {
+	opts := &gitlab.ListUsersOptions{Username: gitlab.String(name)}
+	users, _, err := apiClient.Users.ListUsers(opts)
+	if err != nil {
+		return nil, err
+	}
+
+	if len(users) != 1 {
+		return nil, errors.New(fmt.Sprintf("failed to find user by name : %s", name))
```
profclems: I would prefer you use `fmt.Errorf()` instead of having `errors.New(fmt.Sprintf())` to reduce complexity

profclems (changes requested):

pkg/api/merge_request.go
```diff
@@ -52,6 +52,22 @@ var ListMRs = func(client *gitlab.Client, projectID interface{}, opts *gitlab.Li
...
 }
 
+var ListMRsWithAssignees = func(client *gitlab.Client, projectID interface{}, opts *gitlab.ListProjectMergeRequestsOptions, assigneeIds []int) ([]*gitlab.MergeRequest, error) {
+	if client == nil {
+		client = apiClient
+	}
+	mrs := make([]*gitlab.MergeRequest, 0)
+	for _, id := range assigneeIds {
+		opts.AssigneeID = gitlab.Int(id)
+		assingeMrs, _, err := client.MergeRequests.ListProjectMergeRequests(projectID, opts)
```
profclems: Since we already have `ListMRs` function, why don't you just call that with `ListMRs(client, projectID, opts)`  instead of `client.MergeRequests.ListProjectMergeRequests(projectID, opts)`. 
I know this whole `api` package is a silly wrapper but I'll prefer to keep it that way in case we want to switch from the go-gitlab package

pkg/api/user.go
profclems: Oops. just realized you fixed this with cd41e36 👍

profclems:
@zemzale, great work on this PR 👍

pkg/api/merge_request.go
zemzale: Yeah, this makes sense. That is the same thing I am doing for users in `api` package. Changed this.

zemzale: Yeah, this makes sense. That is the same thing I am doing for users in `api` package. Changed this.

profclems (approved):
Thanks @zemzale. LGTM! 🎉

PR #309, opened 2020-11-16
fix(git): remove surrounding quotes title

guyfedwards:
**Description**

The return value from `--format='%h %s'` includes single quotes, which
seem not to be needed and causing the issue so this commit removes them

**Related Issue**

**Motivation and Context**

**How Has This Been Tested?**

image

**Types of changes**

**Checklist:**

- [x] All new and existing tests passed.


internal/git/git.go
```diff
@@ -150,7 +150,7 @@ type Commit struct {
 }
 
 func LatestCommit(ref string) (*Commit, error) {
-	logCmd := GitCommand("show", "-s", "--format='%h %s'", ref)
+	logCmd := GitCommand("show", "-s", "--format=%h %s", ref)
```
profclems, line 153: Thanks for spotting this problem but can you test your solution and confirm that it works?

Because the quotes you removed doesn't seem to be the problem here. It is likely to error because instead of `git show -s --format='%h %s' ref`, it will rather be `git show -s --format=%h %s ref` where the value for `--format` flag will be without quotes (i.e. `--format=%h %s` instead of `--format='%h %s'`) and ***may*** generate this error when run
```sh
$ git show -s --format=%h %s  master
fatal: ambiguous argument '%s': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
```

guyfedwards: Yea I thought this too but Go seems to format it correctly and escape the space. I have tested on my machine but that is linux so possible it could cause issues elsewhere. 

Another easy fix would be to just strip the quotes from the output. Let me see if I can write a test to confirm it works.

guyfedwards: I'm not particularly proficient in writing tests in go so wasn't able to get a worthwhile usecase. I see there is a test git repo set up to run tests against which may allow this to be validated but I coulnd't quite figure it out. I did test it against a test repo i have and can confirm it certainly works for me, again on linux.

profclems: Great! My only concern is if it will work on windows

profclems: For the test case, there's a test for mr_create that tests the autofill as well. You could change this line:
https://github.com/profclems/glab/blob/1666742ac2b17b12a3288b13f619e87dd6208954/commands/mr/create/mr_create_test.go#L130

to be:
```go
cmdtest.Eq(t, out, "!1 Update somefile.txt")
```

This will expect the line to be exactly equal to `!1 Update somefile.txt`. So if the output is `!1 Update somefile.txt'` the test will fail accordingly

profclems: Run the test with `make test`

guyfedwards: Nice, thanks. I amended the test. Ended up sticking with the `assert.Contains` as `out` had a bunch of other lines in there so seemed cleaner than `strings.Split(out)[1]` etc. 

I'll open an issue but currently the tests partially fail on my machine, it looks like this is to do with the fact some are looking explicitly for `profclems/glab` in the git remote which for my fork doesn't exist, and also for the `TestNewCmdCreate_autofill` test, it eems to try and push the branch, which again fails.

profclems: Yhyhyh... You're right. I almost forgot that. I need to fix that.

By the way, the test passed on the CI.

profclems: I have fixed the test issue with #317 

I have delayed merging this because I am unsure whether it will work on windows or not. I am adding windows test to the CI with #314

guyfedwards: Nice one

profclems: This works on windows as well. Thanks

PR #328, opened 2020-11-23
feat(api): add api command to allow access to GitLab API

profclems:
**Description**
This PR adds `api` command to `glab` core commands to allow users make authenticated requests to the GitLab API v4 endpoints and GitLab GraphQL endpoints.

It prints a pretty formatted JSON response.

**Motivation and Context**

The GitLab CLI tool (`gh`) has the `api` command that allows user to easily make authenticated requests to GitHub. 

This is useful for cases where there are no commands available to perform specific tasks.

**How Has This Been Tested?**

Added a couple of tests and also manually tested as shown in the screenshot

Peek 2020-11-23 12-21

**Types of changes**


commands/api/api.go
```diff
@@ -0,0 +1,543 @@
...
+		return value, err
+	}
+
+	filled := placeholderRE.ReplaceAllStringFunc(value, func(m string) string {
+		switch m {
+		case ":group/:namespace/:repo":
+			return url.PathEscape(baseRepo.FullName())
+		case ":namespace/:repo":
+			return url.PathEscape(baseRepo.RepoNamespace() + "/" + baseRepo.RepoName())
+		case ":fullpath":
```
zemzale: nit: This could be combined with the case `case ":group/:namespace/:repo":` just above

commands/api/api.go
```diff
@@ -0,0 +1,543 @@
...
+		case ":repo":
+			return baseRepo.RepoName()
+		case ":branch":
+			branch, e := opts.Branch()
+			if e != nil {
+				err = e
+			}
+			return branch
+		default:
+			panic(fmt.Sprintf("invalid placeholder: %q", m))
```
zemzale: Maybe we can avoid panicing by just setting `err` here?

commands/api/pagination.go
```diff
@@ -0,0 +1,107 @@
...
+			case '}', ']':
+				stack = stack[:len(stack)-1]
+				contextKey = ""
+				idx = 0
+			}
+		default:
+			isKey := len(stack) > 0 && stack[len(stack)-1] == '{' && idx%2 == 0
+			idx++
+
+			switch tt := t.(type) {
```
zemzale: I would suggest, we maybe flatten this switch.
The part where `isKey` is being calculated t.i. 
```go
isKey := len(stack) > 0 && stack[len(stack)-1] == '{' && idx%2 == 0
idx++
```
Then we could just remove the nested switch , and replace the default case above with the cases inside the this switch

commands/api/api.go
```diff
@@ -0,0 +1,543 @@
...
+
+	cmd.Flags().StringVar(&opts.Hostname, "hostname", "", "The GitLab hostname for the request (default is \"gitlab.com\" or authenticated host in current git directory)")
+	cmd.Flags().StringVarP(&opts.RequestMethod, "method", "X", "GET", "The HTTP method for the request")
+	cmd.Flags().StringArrayVarP(&opts.MagicFields, "field", "F", nil, "Add a parameter of inferred type")
+	cmd.Flags().StringArrayVarP(&opts.RawFields, "raw-field", "f", nil, "Add a string parameter")
+	cmd.Flags().StringArrayVarP(&opts.RequestHeaders, "header", "H", nil, "Add an additional HTTP request header")
+	cmd.Flags().BoolVarP(&opts.ShowResponseHeaders, "include", "i", false, "Include HTTP response headers in the output")
+	cmd.Flags().BoolVar(&opts.Paginate, "paginate", false, "Make additional HTTP requests to fetch all pages of results")
+	cmd.Flags().StringVar(&opts.RequestInputFile, "input", "", "The file to use as body for the HTTP request")
+	cmd.Flags().BoolVar(&opts.Silent, "silent", false, "Do not print the response body")
```
zemzale, line 188: Could we maybe add a flag for not pretty printing? This would be usefull for pipeing the output to something else like 
```sh
glab api /projects | jq `.foo`
```

zemzale, line 185: Could we maybe add a flag for not pretty printing? This would be usefull for pipeing the output to something else like 
```sh
glab api /projects | jq `.foo`
```

zemzale (commented):
When locally testing, I had a bad token in my config, and it the pager didn't close untill I pressed `ctrl-c` and there was a message `GET https://gitlab.com/api/v4/projects: 401 {message: 401 Unauthorized}`

commands/api/api.go
profclems: `glab` should automatically detect when pipeing output and on non-ttys and prevent color outputs and pretty-printing the json output. Tested with these and properly working
```sh
$ glab api /projects | tee file.txt

$ glab api /projects > file.txt

$ glab api /projects | jq -c > '.foo'

$ glab api /projects | jq > '.foo'
```

commands/api/api.go
profclems: Yeah you're right. I used panic while testing and forgot to replace it with `err`

commands/api/api.go
zemzale: Nice! Didn't know this.

commands/api/pagination.go
zemzale: I just understood that I wrote this comment kinda leaving out some words. 
I intended to mean we could rewrite this as 
```go
isKey := len(stack) > 0 && stack[len(stack)-1] == '{' && idx%2 == 0
idx++
switch tt := t.(type) {
...
case string:
...
case bool:
...
}
```
This way we can remove one of the nested switches, but if you think that there is no need for this then we can just leave it as is.

profclems (commented):
In order to have full control over the http response, I had to make non-breaking changes to the api Client since go-gitlab does not return the actual error response sent by the GitLab API.
Should resolve the issue stated by @zemzale in  https://github.com/profclems/glab/pull/328#pullrequestreview-537342472:

> When locally testing, I had a bad token in my config, and it the pager didn't close untill I pressed ctrl-c and there was a message GET https://gitlab.com/api/v4/projects: 401 {message: 401 Unauthorized}

commands/api/pagination.go
profclems: Yeah got it... Resolved now

zemzale:
Tested localy and works and now I can indeed see that I am not authorized.

profclems:
Great!

PR #336, opened 2020-11-28
fix(ci): Remove OrderBy field from pipeline list options

### LINKED ISSUE #335: [Bug] GitLab 11.11 failure to run `glab pipeline status -l`
**Description**
Opening this issue to track it separately from #325 since it's issue with the same command, but is not the same issue. 
As per comments https://github.com/profclems/glab/issues/325#issuecomment-731756008 https://github.com/profclems/glab/issues/325#issuecomment-731827690 there is failure to run `glab pipeline status -l` on `GitLab 11.11`.
This is because we use `order_by:updated_at` and it [wasn't in the GitLab 11.11 API](https://docs.gitlab.com/11.11/ee/api/pipelines.html#list-project-pipelines)
**Expected Behavior**
Should be execute the command and view pipeline status
**Actual Behavior**
The response is :
```sh
GET https://gitlab..../api/v4/projects/tools/templates/pipelines: 400 {error: order_by does not have a valid value}
```
**Possible Fix**

We could check the version of the GitLab we are running against, and if it's 11.11 then we just don't pass this order by parameter. 
I have not looked into how important is the ordering but if it's needed we can just request these pipelines one by one and order them ourselfs. 
**Steps to Reproduce**

1. Try to execute `glab pipeline status -l` on GitLab 11.11

**Your Environment**

As per https://github.com/profclems/glab/issues/325#issuecomment-731827690
* Version used: `GitLab 11.11` `glab 1.11.1` (note: `GitLab 12.10` is mentioned too, but that shouldn't be affected)

zemzale:
**Description**
This field is not supported in GitLab 11.11. This doesn't allow this
command to be run against this version of GitLab.

Since this ordering is not that important easiest is just to remove it.

See: https://docs.gitlab.com/11.11/ee/api/pipelines.html#list-project-pipelines

**Related Issue**

Fixes #335

**Types of changes**

**Checklist:**

- [x] All new and existing tests passed.

### DIFF
--- a/commands/pipeline/ci/trace/pipeline_ci_trace.go
+++ b/commands/pipeline/ci/trace/pipeline_ci_trace.go
@@ -61,9 +61,8 @@ func TraceCmdFunc(cmd *cobra.Command, args []string, f *cmdutils.Factory) error 
 		jobID = utils.StringToInt(args[0])
 	} else {
 		l := &gitlab.ListProjectPipelinesOptions{
-			Ref:     gitlab.String(branch),
-			OrderBy: gitlab.String("updated_at"),
-			Sort:    gitlab.String("desc"),
+			Ref:  gitlab.String(branch),
+			Sort: gitlab.String("desc"),
 		}
 
 		l.Page = 1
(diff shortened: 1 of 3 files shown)


profclems (approved):
Great thanks

PR #394, opened 2020-12-23
fix(internal/glrepo/resolver): don't use pointers to gitlab.Project{}

### LINKED ISSUE #385: glab asks which repository when multiple remotes are present but uses 'origin' repository anyway
**Description**
Screenshot should explain it all, It asks if I want Leo/aports or alpine/aports, I pick the latter but the mr view command still uses the old one.

image

It to check against the picked repo after asking me a repo to pick

**Steps to Reproduce**
1. Clone a gitlab repo from gitlab.alpinelinux.org like gitlab.alpinelinux.org/Leo/atools
2. Add a second remote
3. try to run any command querying a merge request/issue
4. when it asks which one you want to query in by default answer the one that isn't origin
5. watch it use the one that is origin

**Your Environment**

* Version used (Run `glab --version`): glab version 1.12.1-82-g0c3ff7e (2020-12-21)
* Operating System and version: Fedora SilverBlue 33

maxice8:
**Description**

Using pointers trigger a very interesting bug, imagine this case:

Remotes:
- origin (git@gitlab.alpinelinux.org:Leo/aports.git)
- upstream (git@gitlab.alpinelinux.org:alpine/aports.git)

More important information:
- Leo/aports is a fork of alpine/aports.

What happens ?

This piece of code runs, in this case r.network has the info from the 2
remotes.

```go
for _, repo := range r.network {
    if repo.ForkedFromProject != nil {
        fProject, _ := api.GetProject(r.apiClient, repo.ForkedFromProject.PathWithNamespace)
        add(fProject)
    }
    add(&repo)
}
```

Here is the add function that will be referenced:

```go
add := func(r *gitlab.Project) {
    fn, _ := FullNameFromURL(r.HTTPURLToRepo)
    if _, ok := repoMap[fn]; !ok {
        repoMap[fn] = r
        repoNames = append(repoNames, fn)
    }
}
```

1. upstream (alpine/aports) is checked
2. it is not a fork so just pass &repo to the add function which adds
   it to repoMap[alpine/aports] which will point to the reference
   (due to &repo) and repoNames
3. origin (Leo/aports) is checked
4. origin (Leo/aports) is a fork!, check for what it is forked from (alpine/aports)
5. pass the fork origin to add, which does nothing because we already
   added repoMap[alpine/aports] and repoNames from the upstream remote
6. pass &repo to the add function which adds it to repoMap[Leo/aports]
   and repoNames

both repoMap[alpine/aports] and repoMap[Leo/aports] now point to the
same address, and the later one (Leo/aports), overwrote the earlier one
(alpine/aports)!

**Related Issue**

Resolves #385

**How Has This Been Tested?**
Tested it against the usecase presented in #385

**Types of changes**

- [ ] Documentation

### DIFF
--- a/internal/glrepo/resolver.go
+++ b/internal/glrepo/resolver.go
@@ -93,7 +93,12 @@ func (r *ResolvedRemotes) BaseRepo(prompt bool) (Interface, error) {
 	add := func(r *gitlab.Project) {
 		fn, _ := FullNameFromURL(r.HTTPURLToRepo)
 		if _, ok := repoMap[fn]; !ok {
-			repoMap[fn] = r
+			// This is run inside a for-loop, create a local copy and use its address
+			// instead of the one given to us, otherwise the value in repoMap will be
+			// overwriten in the next iteration of the loop
+			// See: #395, #384
+			localCopy := *r
+			repoMap[fn] = &localCopy
 			repoNames = append(repoNames, fn)
 		}
 	}


profclems:
@maxice8 would you be available to help maintain glab? I have been a bit tight this week since I'm interviewing for a new role. Let me know if you are ready to join glab's core contributors so I sent you a collaborator invite

profclems (commented):
Makes sense! 
Maybe there should be a better way to fix this.
I'm actually thinking of the cost of copying `gitlab.Project` which is a complex structure than copying the address 🤔. This will be quite a tradeoff

maxice8:
> would you be available to help maintain glab?

sure

maxice8:
Here is a stackoverflow talking about it: https://stackoverflow.com/questions/48826460/using-pointers-in-a-for-loop

Apparently this is part of the documentation spec, I guess we can modify `add` so it makes a local copy instead of storing a pointer

maxice8:
Here is a different fix: (build on top of the current one)

```diff
diff --git a/internal/glrepo/resolver.go b/internal/glrepo/resolver.go
index 5c0a627..a1b5a02 100644
--- a/internal/glrepo/resolver.go
+++ b/internal/glrepo/resolver.go
@@ -89,11 +89,12 @@ func (r *ResolvedRemotes) BaseRepo(prompt bool) (Interface, error) {
 	}
 
 	var repoNames []string
-	repoMap := map[string]gitlab.Project{}
-	add := func(r gitlab.Project) {
+	repoMap := map[string]*gitlab.Project{}
+	add := func(r *gitlab.Project) {
 		fn, _ := FullNameFromURL(r.HTTPURLToRepo)
 		if _, ok := repoMap[fn]; !ok {
-			repoMap[fn] = r
+			localCopy := *r
+			repoMap[fn] = &localCopy
 			repoNames = append(repoNames, fn)
 		}
 	}
@@ -101,9 +102,9 @@ func (r *ResolvedRemotes) BaseRepo(prompt bool) (Interface, error) {
 	for _, repo := range r.network {
 		if repo.ForkedFromProject != nil {
 			fProject, _ := api.GetProject(r.apiClient, repo.ForkedFromProject.PathWithNamespace)
-			add(*fProject)
+			add(fProject)
 		}
-		add(repo)
+		add(&repo)
 	}
 
 	if len(repoNames) == 0 {
```

Keep using pointers but in the add function instead of using the pointer given to us, if it is applicable we make a copy of the structure and use the pointer to it instead.

The advantage is that we won't need copying unless we are **certain** that we will be storing it in repoMap

profclems:
> Here is a different fix: (build on top of the current one)
>  ....
> Keep using pointers but in the add function instead of using the pointer given to us, if it is applicable we make a copy of the structure and use the pointer to it instead.
> 
> The advantage is that we won't need copying unless we are **certain** that we will be storing it in repoMap

Good one. Does it still work when tested against the use-case presented here #385?

maxice8:
> > Here is a different fix: (build on top of the current one)
> > ....
> > Keep using pointers but in the add function instead of using the pointer given to us, if it is applicable we make a copy of the structure and use the pointer to it instead.
> > The advantage is that we won't need copying unless we are **certain** that we will be storing it in repoMap
> 
> Good one. Does it still work when tested against the use-case presented here #385?

Yes, if it didn't then there would be no point in proposing it in the first place.

I pushed this suggested version with a comment, please check it.

profclems (approved):

profclems:
I'm okay with this approach

PR #396, opened 2020-12-23
feat(cmd/issue/update): add --public and --confidential flags

### LINKED ISSUE #395: Allow making an issue confidential/public
Alpine Linux uses confidential issues to discuss sensitive topics like upcoming security fixes, after the security issue is fixed or deemed not enough to warrant a fix these issues are made public

**Propose a Solution**
add a `--confidential={yes|no}` switch to `glab issue update`

**Additional context**
Something akin to what I do on [agl](https://gitlab.alpinelinux.org/Leo/agl) on [cmd/issue/edit.go#L120](https://gitlab.alpinelinux.org/Leo/agl/-/blob/master/cmd/issue/edit.go#L120)

image

maxice8:
**Description**

they allow making an issue public or confidential, as documented by
GitLab here[1] and in their API[2] (see confidential field)

[1]: https://docs.gitlab.com/ee/user/project/issues/confidential_issues.html
[2]: https://docs.gitlab.com/ee/api/issues.html#edit-issue

**Related Issue**

Resolves #395 

**How Has This Been Tested?**
Tested against the usecase presented in #395 

**Types of changes**

- [ ] Documentation

### DIFF
--- a/commands/issue/update/issue_update.go
+++ b/commands/issue/update/issue_update.go
@@ -1,6 +1,7 @@
 package update
 
 import (
+	"errors"
 	"fmt"
 
 	"github.com/profclems/glab/commands/cmdutils"
@@ -55,6 +56,16 @@ func NewCmdUpdate(f *cmdutils.Factory) *cobra.Command {
 				l.RemoveLabels = gitlab.Labels(m)
 			}
 
+			if cmd.Flags().Changed("confidential") && cmd.Flags().Changed("public") {
+				return &cmdutils.FlagError{Err: errors.New("--public and --confidential can't be used together")}
+			}
+			if m, _ := cmd.Flags().GetBool("public"); m {
+				l.Confidential = gitlab.Bool(false)
+			}
+			if m, _ := cmd.Flags().GetBool("confidential"); m {
+				l.Confidential = gitlab.Bool(true)
+			}
+
 			fmt.Fprintf(out, "- Updating issue #%d\n", issueID)
 
 			issue, err := api.UpdateIssue(apiClient, repo.FullName(), issueID, l)
@@ -74,6 +85,8 @@ func NewCmdUpdate(f *cmdutils.Factory) *cobra.Command {
 	issueUpdateCmd.Flags().StringP("description", "d", "", "Issue description")
 	issueUpdateCmd.Flags().StringArrayP("label", "l", []string{}, "add labels")
 	issueUpdateCmd.Flags().StringArrayP("unlabel", "u", []string{}, "remove labels")
+	issueUpdateCmd.Flags().BoolP("public", "p", false, "Make issue public")
+	issueUpdateCmd.Flags().BoolP("confidential", "c", false, "Make issue confidential")
 
 	return issueUpdateCmd
 }


commands/issue/update/issue_update.go
```diff
@@ -74,6 +80,8 @@ func NewCmdUpdate(f *cmdutils.Factory) *cobra.Command {
 	issueUpdateCmd.Flags().StringP("description", "d", "", "Issue description")
 	issueUpdateCmd.Flags().StringArrayP("label", "l", []string{}, "add labels")
 	issueUpdateCmd.Flags().StringArrayP("unlabel", "u", []string{}, "remove labels")
+	issueUpdateCmd.Flags().BoolP("public", "", false, "Make issue public")
```
profclems: Add `p` as a shorthand for `public`
```suggestion
	issueUpdateCmd.Flags().BoolP("public", "p", false, "Make issue public")
```

commands/issue/update/issue_update.go
```diff
@@ -74,6 +80,8 @@ func NewCmdUpdate(f *cmdutils.Factory) *cobra.Command {
...
+	issueUpdateCmd.Flags().BoolP("confidential", "", false, "Make issue confidential, overrides --public")
```
profclems: Add `c` as a shorthand for `confidential`

```suggestion
	issueUpdateCmd.Flags().BoolP("confidential", "c", false, "Make issue confidential, overrides --public")
```

profclems (commented):
Thanks @maxice8

maxice8:
Is there some documentation on how to add flags ? In agl I just have `--confidential` which is a string that can take {yes|no}, same with `--discussion={locked|unlocked}`

profclems (approved):

profclems:
> In agl I just have `--confidential` which is a string that can take {yes|no}, same with `--discussion={locked|unlocked}`

Cobra doesn't have any built-in function to restrict flag values so you would have to validate it yourself. 
Have a look at this issue https://github.com/spf13/cobra/issues/1006

maxice8:
> > In agl I just have `--confidential` which is a string that can take {yes|no}, same with `--discussion={locked|unlocked}`
> 
> Cobra doesn't have any built-in function to restrict flag values so you would have to validate it yourself.
> Have a look at this issue [spf13/cobra#1006](https://github.com/spf13/cobra/issues/1006)

Yes, in agl I just check with `switch case` for the correct values and never allow them to be passed wrongly, here I decided to let one override the other. 

If one passes `--confidential` and `--public` then we err on the side of safety and `--confidential` will override `--public`

This is expressed in the code by the logic for the flag that can override being later so the condition for `--public` can set `l.Confidentiality = gitlab.Bool(false)` but it won't matter because the logic for `--confidential` will trigger later and `l.Confidentiality = gitlab.Bool(true)`

---

What I actually mean is there a guideline on new flags ? Should we have different `--public` and `--confidential` flags that act on the same attribute of a GitLab object ?

profclems:
> What I actually mean is there a guideline on new flags ? Should we have different `--public` and `--confidential` flags that act on the same attribute of a GitLab object ?

You can but you should throw an error when both flags are specified. You can do this check

```go
if cmd.Flags().Changed("confidential") && cmd.Flags().Changed("public") {
	return &cmdutil.FlagError{Err: errors.New("specify either of --public or --confidential")}
}
```

Sorry, I think I approved this too soon. This check should be added instead of having one override the other. This should be safe.

commands/issue/update/issue_update.go
```diff
@@ -54,6 +54,12 @@ func NewCmdUpdate(f *cmdutils.Factory) *cobra.Command {
 			if m, _ := cmd.Flags().GetStringArray("unlabel"); len(m) != 0 {
 				l.RemoveLabels = gitlab.Labels(m)
 			}
+			if m, _ := cmd.Flags().GetBool("public"); m {
+				l.Confidential = gitlab.Bool(false)
+			}
+			if m, _ := cmd.Flags().GetBool("confidential"); m {
+				l.Confidential = gitlab.Bool(true)
+			}
```
profclems, line 67: Instead of one overriding the other, there should be an error thrown if both flags are specified

```go
if cmd.Flags().Changed("confidential") && cmd.Flags().Changed("public") {
    return &cmdutil.FlagError{Err: errors.New("specify either of --public or --confidential")}
}
```

profclems (changes requested):

maxice8:
image

profclems (approved):

PR #404, opened 2020-12-23
feat(commands/mr/create): add --repo support for --target-project

### LINKED ISSUE #403: Support for --repo specification into --target-project in mr create
Instead of forcing the users to get the ID themselves, allow them to use the specification just like `--repo` for the repo

**Propose a Solution**
make `--target-project` take a string instead

maxice8:
**Description**

one can now use something like `--target-project alpine/aports`
to set the target of the project something via a string, it uses
glrepo.FromFullName just like --repo so it supports the same
specification.

It also re-uses the Httpclient from mr so it will use the correct HOST
(as long as GitLab does not become federated we will be fine in this
case)

**Related Issue**

Resolves #403 

**How Has This Been Tested?**

Created https://gitlab.alpinelinux.org/alpine/aports/-/merge_requests/16137 with `glab mr create --target-project alpine/aports`

image
image

**Types of changes**

- [ ] Documentation

### DIFF
--- a/commands/mr/create/mr_create.go
+++ b/commands/mr/create/mr_create.go
@@ -3,6 +3,7 @@ package create
 import (
 	"errors"
 	"fmt"
+	"strconv"
 	"strings"
 
 	"github.com/profclems/glab/internal/config"
@@ -51,6 +52,8 @@ type CreateOpts struct {
 }
 
 func NewCmdCreate(f *cmdutils.Factory) *cobra.Command {
+	var mrCreateTargetProject string
+
 	opts := &CreateOpts{
 		IO:       f.IO,
 		Branch:   f.Branch,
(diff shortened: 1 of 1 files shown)


maxice8:
This is a proof of concept of what can be done, there are lots of optimizations that can be done, as we currently just do lots of conversion with `strconv.Itoa` and `strconv.Atoi`

profclems:
> This is a proof of concept of what can be done, there are lots of optimizations that can be done, as we currently just do lots of conversion with `strconv.Itoa` and `strconv.Atoi`

Yeah I get it.

Let me know when this PR is ready

maxice8:
> > This is a proof of concept of what can be done, there are lots of optimizations that can be done, as we currently just do lots of conversion with `strconv.Itoa` and `strconv.Atoi`
> 
> Yeah I get it.
> 
> Let me know when this PR is ready

The point of the proof of concept is to discuss if it is really the best path and if yes how to improve it.

Unless you're giving me  `carte blanche` to implement how I see fit

profclems (approved):
This implementation is much better than the former. At least it prevents doing a lot of conversions with `strconv.Atoi` and `strconv.Itoa`

maxice8:
It should be good to go, but mr_create as a whole will be subject to a rewrite in the future (possibly for a new major version) to support things like `--rebase`, `glab-resolved = head` and `glab-resolved = target`

PR #410, opened 2020-12-25
glab-resolved = {base|head} support

maxice8:
**Description**

This is a big one:

- Create HeadRepo in internal/glrepo, works the same as `BaseRepo` but checks `glab-resolved = head` and
  also picks the first forked repo instead of just the first repo
- vendor functions used for `glab-resolved = base` into create_mr and make them work for `glab-resolved = head`
- deprecate `--target-project` as users can just use `--repo` instead

**Related Issue**

Resolves #406 #393 

**How Has This Been Tested?**

Alpine

**Types of changes**

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation
- [ ] Chore (Related to CI or Packaging to platforms)

**TODO**

- Need to re-introduce the alternate code for resolution but make it prefix with `base:` and `head:`

PR #414, opened 2020-12-26
Rework BaseRepo from internal.glrepo.resolver

maxice8:
**Description**

This reworks BaseRepo to support certain features:

- Be stricter about cached resolutions, only use `base` or the first with the prefix `base:` (instead of any other resolution)

**Related Issue**

No issue was open for this but it is ripped of #410 to be easier to reason about

**How Has This Been Tested?**

Tested as part of #410 

**Types of changes**

- [ ] Documentation

### DIFF
--- a/internal/glrepo/resolver.go
+++ b/internal/glrepo/resolver.go
@@ -117,12 +117,12 @@ func (r *ResolvedRemotes) BaseRepo(prompt bool) (Interface, error) {
 		}
 	}
 
-	for _, repo := range r.network {
-		if repo.ForkedFromProject != nil {
-			fProject, _ := api.GetProject(r.apiClient, repo.ForkedFromProject.PathWithNamespace)
+	for i := range r.network {
+		if r.network[i].ForkedFromProject != nil {
+			fProject, _ := api.GetProject(r.apiClient, r.network[i].ForkedFromProject.PathWithNamespace)
 			add(fProject)
 		}
-		add(&repo)
+		add(&r.network[i])
 	}
 
 	if len(repoNames) == 0 {
--- a/internal/glrepo/resolver.go
+++ b/internal/glrepo/resolver.go
@@ -3,6 +3,7 @@ package glrepo
 import (
 	"errors"
 	"sort"
+	"strings"
 
 	"github.com/AlecAivazis/survey/v2"
 	"github.com/profclems/glab/internal/git"
@@ -66,7 +67,7 @@ func (r *ResolvedRemotes) BaseRepo(prompt bool) (Interface, error) {
 	for _, r := range r.remotes {
 		if r.Resolved == "base" {
 			return r, nil
-		} else if r.Resolved != "" {
+		} else if strings.HasPrefix(r.Resolved, "base:") {
 			repo, err := FromFullName(r.Resolved)
 			if err != nil {
 				return nil, err
--- a/internal/glrepo/resolver.go
+++ b/internal/glrepo/resolver.go
@@ -68,10 +68,23 @@ func (r *ResolvedRemotes) BaseRepo(prompt bool) (Interface, error) {
 		if r.Resolved == "base" {
 			return r, nil
 		} else if strings.HasPrefix(r.Resolved, "base:") {
+			repo, err := FromFullName(strings.TrimPrefix(r.Resolved, "base:"))
+			if err != nil {
+				return nil, err
+			}
+			return NewWithHost(repo.RepoOwner(), repo.RepoName(), r.RepoHost()), nil
+		} else if r.Resolved != "" && !strings.HasPrefix(r.Resolved, "head:") {
+			// Backward compatibility kludge for remoteless resolutions created before
+			// BaseRepo started creeating resolutions prefixed with `base:`
 			repo, err := FromFullName(r.Resolved)
 			if err != nil {
 				return nil, err
 			}
+			// Rewrite resolution, ignore the error as this will keep working
+			// in the future we might add a warning that we couldn't rewrite
+			// it for compatiblity
+			_ = git.SetRemoteResolution(r.Name, "base:"+r.Resolved)
+
 			return NewWithHost(repo.RepoOwner(), repo.RepoName(), r.RepoHost()), nil
 		}
 	}
--- a/internal/glrepo/resolver.go
+++ b/internal/glrepo/resolver.go
@@ -107,12 +107,7 @@ func (r *ResolvedRemotes) BaseRepo(prompt bool) (Interface, error) {
 	add := func(r *gitlab.Project) {
 		fn, _ := FullNameFromURL(r.HTTPURLToRepo)
 		if _, ok := repoMap[fn]; !ok {
-			// This is run inside a for-loop, create a local copy and use its address
-			// instead of the one given to us, otherwise the value in repoMap will be
-			// overwriten in the next iteration of the loop
-			// See: #395, #384
-			localCopy := *r
-			repoMap[fn] = &localCopy
+			repoMap[fn] = r
 			repoNames = append(repoNames, fn)
 		}
 	}
--- a/internal/glrepo/resolver.go
+++ b/internal/glrepo/resolver.go
@@ -148,6 +148,7 @@ func (r *ResolvedRemotes) BaseRepo(prompt bool) (Interface, error) {
 	if remote == nil {
 		remote = r.remotes[0]
 		resolution, _ = FullNameFromURL(selectedRepo.HTTPURLToRepo)
+		resolution = "base:" + resolution
 	}
 
 	// cache the result to git config


maxice8:
@profclems I have split out the part that touches `BaseRepo` in `internal/glrepo/resolver.go` into this as it is quite independent from the rest of #410

internal/glrepo/resolver.go
```diff
@@ -134,6 +130,7 @@ func (r *ResolvedRemotes) BaseRepo(prompt bool) (Interface, error) {
 	if remote == nil {
 		remote = r.remotes[0]
 		resolution, _ = FullNameFromURL(selectedRepo.HTTPURLToRepo)
+		resolution = "base:" + resolution
```
maxice8, line 146: This is a part that breaks @profclems

internal/glrepo/resolver.go
```diff
@@ -66,7 +67,7 @@ func (r *ResolvedRemotes) BaseRepo(prompt bool) (Interface, error) {
 	for _, r := range r.remotes {
 		if r.Resolved == "base" {
 			return r, nil
-		} else if r.Resolved != "" {
+		} else if strings.HasPrefix(r.Resolved, "base:") {
```
maxice8, line 70: same here, we need to distinguish resolutions made for `head:` and for `base:`

profclems: Oh okay. I get it

internal/glrepo/resolver.go
profclems: I'm thinking we can get the older `glab-resolved=base` if set and rename it

maxice8: From what I understand these resolutions (the ones that are not `base`  or `head`) are made because there is no matching remote.

I think we can avoid breakage if for `BaseRepo` we check for `glab-resolved = base` and the first `glab-resolved` that does not have the value of `head:`

Resolutions made without the prefix continue to work that way.

maxice8: so if `strings.HasPrefix(r.Resolved, "base:")`  fails we fallback on trying `!strings.hasPrefix(r.Resolved, "head:")` which will match those before

internal/glrepo/resolver.go
```diff
@@ -66,11 +67,24 @@ func (r *ResolvedRemotes) BaseRepo(prompt bool) (Interface, error) {
...
 			repo, err := FromFullName(r.Resolved)
 			if err != nil {
 				return nil, err
 			}
+			return NewWithHost(repo.RepoOwner(), repo.RepoName(), r.RepoHost()), nil
+		} else if !strings.HasPrefix(r.Resolved, "head:") {
```
maxice8: This block makes the whole thing non-breaking.

internal/glrepo/resolver.go
maxice8: See: https://github.com/profclems/glab/pull/414/files#r549038702

profclems:
I checked out to this PR and tried running by first adding another remote and removing the `glab-resolved` from git/config

```sh
ProfClems:  ~/.../profclems/glab  |rework-base-repo {1} ✓|
→ git config -l | grep glab
remote.origin.url=git@github.com:profclems/glab.git
remote.gitlab.url=git@gitlab.com:profclems/glab.git
remote.gitlab.glab-resolved=base

ProfClems:  ~/.../profclems/glab  |rework-base-repo {1} ✓|
→ git remote add gitlab-test git@gitlab.com:glab-cli/test.git

ProfClems:  ~/.../profclems/glab  |rework-base-repo {1} ✓|
→ git config --unset remote.gitlab.glab-resolved

ProfClems:  ~/.../profclems/glab  |rework-base-repo {1} ✓|
→ git config -l | grep glab
remote.origin.url=git@github.com:profclems/glab.git
remote.gitlab.url=git@gitlab.com:profclems/glab.git

ProfClems:  ~/.../profclems/glab  |rework-base-repo {1} ✓|
→ make run run="issue list"
go run -trimpath -ldflags "-X main.build=2020-12-27  -X main.version=v1.12.1-119-g612f631 -X main.debugMode=true" ./cmd/glab issue list
expected the "[HOST/]OWNER/[NAMESPACE/]REPO" format, got ""
```

maxice8:
Fixed, we now check if r.Resolved isn't empty before checking if it doesn't match the prefix

profclems:
Nice! works fine now but this was saved without the `base:` prefix 🤔

```sh
ProfClems:  ~/.../profclems/glab  |rework-base-repo {1} ✓|
→ make run run="issue list"
go run -trimpath -ldflags "-X main.build=2020-12-27  -X main.version=v1.12.1-123-gce44af2 -X main.debugMode=true" ./cmd/glab issue list
? Which should be the base repository (used for e.g. querying issues) for this directory? profclems/glab

ProfClems:  ~/.../profclems/glab  |rework-base-repo {1} ✓|
→ git config -l | grep glab-resolved
remote.gitlab.glab-resolved=base
```

maxice8:
That is working as intended as you picked the repo, it will save on the correct remote

maxice8:
The `base:` and `head:` prefix are used when there is no remote matching so it stores the URL in the first remote it finds  with the prefix + URL

maxice8:
I never used that path of the code but I assume it is used in the following situation.

```txt
$ git remote -v
origin	git@gitlab.alpinelinux.org:Leo/aports.git (fetch)
origin	git@gitlab.alpinelinux.org:Leo/aports.git (push)
$ glab issue list
Which should be the base repository (used for e.g. querying issues) for this directory?
alpine/aports
Leo/aports
$ Which should be the base repository (used for e.g. querying issues) for this directory? alpine/aports
```

In this case `alpine/aports` was added automatically because `Leo/aports` is forked from it:

```golang
for i := range r.network {
	if r.network[i].ForkedFromProject != nil {
		fProject, _ := api.GetProject(r.apiClient, r.network[i].ForkedFromProject.PathWithNamespace)
		add(fProject)
	}
	add(&r.network[i])
}
```

But it has no remote representing it locally, so it saves on the first remote with the full URL.

maxice8:
image

profclems:
Yhyh... I forgot I wasn't testing on a fork.

I just tested on a fork. I added a fork remote and it works fine

```sh
ProfClems:  ~/.../profclems/glab  |rework-base-repo {1} ✓|
→ make run run="issue list"
go run -trimpath -ldflags "-X main.build=2020-12-27  -X main.version=v1.12.1-123-gce44af2 -X main.debugMode=true" ./cmd/glab issue list
? Which should be the base repository (used for e.g. querying issues) for this directory? glab-cli/test


ProfClems:  ~/.../profclems/glab  |rework-base-repo {1} ✓|
→ git config -l | grep glab-resolved
remote.gitlab.glab-resolved=base:glab-cli/test

```

profclems:
One more issue here is that the `base:` is not trimmed from the repo name

```sh
ProfClems:  ~/.../profclems/test  |some-branch ✓|
→ glab issue list
GET https://gitlab.com/api/v4/projects/base:glab-cli/test/issues: 404 {message: 404 Project Not Found}
```

I'm basically reviewing this PR from a user perspective: a functional test kindof 😅

maxice8:
Pushed `strings.Replace(r.Resolved, "base:", "", -1)` please see if it works now

maxice8:
derp, put the replacing in the wrong block

profclems (commented):
I'm curious... Why are you using `Replace()` instead of `TrimPrefix()`. It works though

```go
repo, err := FromFullName(strings.TrimPrefix(r.Resolved, "base:"))
```

Besides if you use `Replace` with `n = -1`, then you could have used `ReplaceAll` instead. But `TrimPrefix` should be faster

maxice8:
> I'm curious... Why are you using `Replace()` instead of `TrimPrefix()`. It works though
> 
> ```go
> repo, err := FromFullName(strings.TrimPrefix(r.Resolved, "base:"))
> ```
> 
> Besides if you use `Replace` with `n = -1`, then you could have used `ReplaceAll` instead. But `TrimPrefix` should be faster

First thing that came to mind

profclems:
> derp, put the replacing in the wrong block

From what I'm seeing, it's in the right block

```go
                } else if strings.HasPrefix(r.Resolved, "base:") {
			repo, err := FromFullName(strings.Replace(r.Resolved, "base:", "", -1))
			if err != nil {
				return nil, err
			}
			return NewWithHost(repo.RepoOwner(), repo.RepoName(), r.RepoHost()), nil
		}
```

maxice8:
> > derp, put the replacing in the wrong block
> 
> From what I'm seeing, it's in the right block
> 
> ```go
>                 } else if strings.HasPrefix(r.Resolved, "base:") {
> 			repo, err := FromFullName(strings.Replace(r.Resolved, "base:", "", -1))
> 			if err != nil {
> 				return nil, err
> 			}
> 			return NewWithHost(repo.RepoOwner(), repo.RepoName(), r.RepoHost()), nil
> 		}
> ```

Because I was fast :rocket:

profclems:
> > > derp, put the replacing in the wrong block
> > 
> > 
> > From what I'm seeing, it's in the right block
> > ```go
> >                 } else if strings.HasPrefix(r.Resolved, "base:") {
> > 			repo, err := FromFullName(strings.Replace(r.Resolved, "base:", "", -1))
> > 			if err != nil {
> > 				return nil, err
> > 			}
> > 			return NewWithHost(repo.RepoOwner(), repo.RepoName(), r.RepoHost()), nil
> > 		}
> > ```
> 
> Because I was fast

😂... You read my mind

profclems (approved):
All looks good now. 

Thanks for your works towards making `glab` better

maxice8:
> All looks good now.
> 
> Thanks for your works towards making `glab` better

No need to thank, I hope I can push `glab` as the endorsed solution for contributing to Alpine Linux (we host our own GitLab instance) from the CLI

PR #427, opened 2020-12-28
fix(ISSUE_TEMPLATE/feature_request): turn explanation into comments

maxice8:
**Description**

Instead of forcing the user to delete these when making a feature
request, turn them into comments instead like it is done for bug reports


profclems (approved):
Makes sense

PR #442, opened 2020-12-29
fix(pkg/api/pipeline): sort pipeline jobs using CreatedAt

### LINKED ISSUE #432: Allow CI pipelines to be displayed with the flow from left to right, rather than right to left
When viewing a pipeline in GitLab's web interface, the first stage is displayed on the left, and the pipeline flows from left to right. In `glab pipeline ci view`, this is reversed.

I find this confusing to navigate, as:

* it differs from what GitLab presents;
* my language (English) runs from left-to-right, so this is the order I'm used to seeing things progress in.

I don't know Go well enough to trace through the code particularly well, but it looks like right-to-left is hardcoded, rather than there being a way to reverse it that I simply haven't found.

**Propose a Solution**
It would be easiest if GLab checked the locale and rendered pipelines left-to-right in locales that use left-to-right languages, and right-to-left in locales that use right-to-left languages; this should automatically do the right thing in most cases.

It might be necessary to override this in some cases (such as where a locale is handled incorrectly) — a `--reverse` flag to `glab pipeline ci view` might be the best option here, although ideally it would be settable through `glab config` so it can be overridden per-repo, per-host, or globally, without the need to add a flag each time.

**Additional context**
My current locale-related environment, which should give left-to-right:

```
LANG=en_GB.UTF-8
LC_ALL=en_GB.UTF-8
LC_CTYPE=en_GB.UTF-8
```

profclems:
**Description**

ListPipelineJobs returns jobs sorted by ID in descending order instead of returning them in the order in which they were created, so we restore the order using the createdAt

**Related Issue**

Resolves #432

**How Has This Been Tested?**

On `glab` [test repo](https://gitlab.com/glab-cli/test)

**On `glab`**
image

**On GitLab**
image

**Types of changes**

- [ ] Documentation

### DIFF
--- a/pkg/api/pipeline.go
+++ b/pkg/api/pipeline.go
@@ -264,26 +264,19 @@ var PipelineJobsWithSha = func(client *gitlab.Client, pid interface{}, sha strin
 			PerPage: 500,
 		},
 	}
-	list, resp, err := client.Jobs.ListPipelineJobs(pid, target, opts)
-	if err != nil {
-		return nil, err
-	}
-	if resp.CurrentPage == resp.TotalPages {
-		return list, nil
-	}
-	opts.Page = resp.NextPage
+	jobsList := make([]*gitlab.Job, 0)
 	for {
 		jobs, resp, err := client.Jobs.ListPipelineJobs(pid, target, opts)
 		if err != nil {
 			return nil, err
 		}
 		opts.Page = resp.NextPage
-		list = append(list, jobs...)
+		jobsList = append(jobsList, jobs...)
 		if resp.CurrentPage == resp.TotalPages {
 			break
 		}
 	}
-	return list, nil
+	return jobsList, nil
 }
 
 var PipelineCILint = func(client *gitlab.Client, content string) (*gitlab.LintResult, error) {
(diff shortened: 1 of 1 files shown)


maxice8:
Can't the `order_by` field be used ?


From the GitLab API  docs here: https://docs.gitlab.com/ee/api/pipelines.html#list-project-pipelines

order_by | string | no | Order pipelines by id, status, ref, updated_at or user_id (default: id)
-- | -- | -- | --

profclems:
> Can't the `order_by` field be used ?
> 
> From the GitLab API docs here: https://docs.gitlab.com/ee/api/pipelines.html#list-project-pipelines
> 
> order_by	string	no	Order pipelines by id, status, ref, updated_at or user_id (default: id)

We get the jobs of that pipeline instead using the [Jobs API](https://docs.gitlab.com/ee/api/jobs.html). It offers no `sort` or `order_by` field.

https://docs.gitlab.com/ee/api/jobs.html

profclems:
Note that `ci view` visually displays the **jobs** for the most recent pipeline of the specified branch

Issue #439, opened 2020-12-28
More interactive options while creating issue/mr


maxice8:
**Describe the feature or problem you'd like to solve**

While creating an issue it would be nicer if we could also set assignees and milestone as we can with the flags

**Propose a Solution**

Prompt the user for adding assignees and milestone just like it is done with labels, for assignees we can list all the users and pick them.

profclems:
I thought of this too, not only for `issue create` but `mr create` as well.

maxice8:
with #446 and #453 all things that can be passed via `--flag` are now supported, except for boolean or uncommon stuff like `--weight`, `--confidential`, `--wip`

PR #446, opened 2020-12-29
add support for interactively attaching milestones

maxice8:
**Description**

This adds:

- Minor support in `pkg/api/milestone.go` (that can be expanded with more stuff as required)
- Helper function like `cmdutils.LabelsPrompt` but for Milestones instead
- Make `mr create` and `issue create` use the the helper function above to interactively attach milestones

**Related Issue**

Partial fix to #439 

**How Has This Been Tested?**

Still undergoing testing

milestone-1
milestone-2
milestone-3

**Types of changes**

- [ ] Documentation

### DIFF
--- a/commands/cmdutils/cmdutils.go
+++ b/commands/cmdutils/cmdutils.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 
 	"github.com/profclems/glab/internal/glrepo"
+	"github.com/profclems/glab/internal/utils"
 	"github.com/profclems/glab/pkg/api"
 	"github.com/xanzy/go-gitlab"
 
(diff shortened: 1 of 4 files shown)


profclems (commented):
I'm so excited about this one. 

I have a few concerns:

- What do we do if a repo has no milestone set? 
  - The current solution still prompts the user if they want to add a milestone, and if the the user selects `yes`, skips if no milestone is found
  - Should we notify the user that there are no milestones available and skip?
  - Or show a prompt where the user can create a new milestone?

maxice8:
tell the user there are no milestones

profclems:
Currently, the user is not notified that there are no milestones, it just skips after the user has selected `yes`

maxice8:
I have reworked it so it will return `There are no milestones in this project`, which is triggered only if the user says `yes`.

commands/cmdutils/cmdutils.go
```diff
@@ -174,6 +174,43 @@ func LabelsPrompt(response *[]string, apiClient *gitlab.Client, repoRemote *glre
...
+	if err != nil {
+		return err
+	}
+	if addMilestones {
+		var milestoneOptions []string
+		milestoneMap := map[string]*gitlab.Milestone{}
+
+		lOpts := &gitlab.ListMilestonesOptions{}
+		lOpts.PerPage = 100
+		milestones, err := api.ListMilestones(apiClient, repoRemote.FullName(), lOpts)
```
profclems: Shouldn't we consider the `state` of the milestones 🤔? This will return both `active` and `closed` milestones

https://docs.gitlab.com/ee/api/milestones.html#list-project-milestones

maxice8: I guess that makes sense, at first I didn't because -m also doesn't discriminate between `active` and `closed` ones, but in an interactive setting it makes sense to cut down on the amount of milestones and have just the ones that are `active`.

In the future if someone asks we can **maybe** add a flag that makes it consider `closed` one too.

Thanks for the insight I'll change it prompty

maxice8: Updated, please check the new code.

profclems: For now, if a user wants to add a closed milestone, they can use `-m` but for the prompt, it's better we cut down the list as you rightly stated

maxice8: Yes I have modifier the opts to include `State: gitlab.String("active")` so the API will only return `active` milestones.

commands/cmdutils/cmdutils.go
```diff
@@ -174,6 +174,46 @@ func LabelsPrompt(response *[]string, apiClient *gitlab.Client, repoRemote *glre
...
+		lOpts := &gitlab.ListMilestonesOptions{
+			State: gitlab.String("active"),
+		}
+		lOpts.PerPage = 100
+		milestones, err := api.ListMilestones(apiClient, repoRemote.FullName(), lOpts)
+		if err != nil {
+			return err
+		}
+		if len(milestones) == 0 {
+			return fmt.Errorf("There are no active milestones in this project")
```
profclems: I'm thinking instead of returning an error which will cause the user to lose info entered, we just print the message to stderr and return nil; this will help skip this step and continue with the submission so there will be no data loss (since there's no way to recover and continue from previous point as there is in gh).

That's my suggestion. What do you think?

If implemented, you may have to change the signature of the function to include the `*utils.IOStreams` in the arguments
```go
func MilestonesPrompt(response *int, apiClient *gitlab.Client, repoRemote *glrepo.Remote, ios *utils.IOStreams) (err error) {
```
Then the error printed with:
```go
fmt.Fprintln(ios.StdErr, "There are no active milestones in this project")
return nil
```

maxice8: Done

profclems (approved):

Issue #405, opened 2020-12-23
"web" flag for more resources


pbnj:
**Describe the feature or problem you'd like to solve**

Similar to `glab mr view -w|--web` functionality, I would like to be able to open other resources in my default browser for most commands.

Something like:

- `glab pipeline -w|--web` to open the pipelines page of the current repo
- `glab repo -w|--web` to open the project page of the current repo
- `glab repo search <term> -w|--web` to open the search results on gitlab
- `glab issue -w|--web` to open the issues page of the current repo
- `glab mr -w|--web` to open the MRs page of the current repo
- `glab release -w|--web` to open the releases page of the current repo

**Propose a Solution**
The logic is already implemented for some APIs, like a single [issue](https://github.com/profclems/glab/blob/d118ca79599a13ae65ef674f0e99d80188cc223e/commands/issue/view/issue_view.go#L68) and a single [mr](https://github.com/profclems/glab/blob/d118ca79599a13ae65ef674f0e99d80188cc223e/commands/mr/view/mr_view.go#L57), so it shouldn't be too difficult to extend this to other resources, like pipelines, repo, repo search, issues, mrs, and releases.

maxice8:
@profclems if `OK` is given then I'll work on this when I get time

pbnj:
I'm also willing to work on this as well.

Just wanted to create the GH issue and get an OK before investing the time/effort on a PR that could get rejected 😅

profclems:
I agree with this!

maxice8:
@pbnj go nuts then :D

profclems:
@pbnj, you still in for it?

pbnj:
@profclems - certainly.

clemsbot:
This issue is stale because it has been open 30 days with no activity. Remove stale label or comment or this will be closed in 5 days.

clemsbot:
This issue has been automatically marked as stale because it has not had recent activity. We haven't had the time to address it yet, but we want to keep it open. This message is just a reminder for us to help triage issues.

PR #447, opened 2020-12-29
Draft: feat(repo): implement web flag

pbnj:
**Description**

Implement `-w|--web` flag on `repo` command to open repo in browser

**Related Issue**

Relates to #405

**How Has This Been Tested?**

- `go build cmd/glab/main.go`
- `./main repo -w`

Screen Shot 2020-12-29 at 3 27 12 PM

**Types of changes**

- [ ] Documentation

### DIFF
--- a/commands/project/repo.go
+++ b/commands/project/repo.go
@@ -1,6 +1,8 @@
 package project
 
 import (
+	"fmt"
+
 	"github.com/profclems/glab/commands/cmdutils"
 	repoCmdArchive "github.com/profclems/glab/commands/project/archive"
 	repoCmdClone "github.com/profclems/glab/commands/project/clone"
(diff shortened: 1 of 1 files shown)


pbnj:
@profclems - this is an initial Draft PR. 

A few notes:

1. Would you like separate PRs for each command?
2. As you can in the screenshot, when user runs the `repo` command without args/flags, there is no output. I need to adjust this to print the help text.

profclems:
> @profclems - this is an initial Draft PR.

Noted! Thanks for taking this up

> 1. Would you like separate PRs for each command?

You can work on that in a single PR. I would like to review them commit by commit

> 2\. As you can in the screenshot, when user runs the `repo` command without args/flags, there is no output. I need to adjust this to print the help text.

I will explain that in a review comment

commands/project/repo.go
```diff
@@ -9,16 +11,40 @@ import (
...
+			}
+			if opts.OpenInBrowser {
+				browser, err := cfg.Get(baseRepo.RepoHost(), "browser")
+				if err != nil {
+					return err
+				}
+				return utils.OpenInBrowser(fmt.Sprintf("https://%s/%s", baseRepo.RepoHost(), baseRepo.FullName()), browser)
+			}
+			return nil
+		},
```
profclems: Instead of using the `RunE`, you might want to consider using a `PreRun` hook instead.
Using the `RunE` will prevent it from executing child commands.

profclems:
```suggestion
		PreRunE: func(cmd *cobra.Command, args []string) error {
			if opts.OpenInBrowser {
				cfg, err := f.Config()
				if err != nil {
					return err
				}
				baseRepo, err := f.BaseRepo()
				if err != nil {
					return err
				}
				browser, err := cfg.Get(baseRepo.RepoHost(), "browser")
				if err != nil {
					return err
				}
				return utils.OpenInBrowser(fmt.Sprintf("https://%s/%s", baseRepo.RepoHost(), baseRepo.FullName()), browser)
			}
			return nil
		},
```

profclems: For the subcommands, you can do that directly in the `RunE`

commands/project/repo.go
```diff
@@ -29,5 +55,7 @@ func NewCmdRepo(f *cmdutils.Factory) *cobra.Command {
 	repoCmd.AddCommand(repoCmdFork.NewCmdFork(f, nil))
 	repoCmd.AddCommand(repoCmdSearch.NewCmdSearch(f))
 
+	repoCmd.Flags().BoolVarP(&opts.OpenInBrowser, "web", "w", false, "Open repo in a browser. Uses default browser or browser specified in BROWSER variable")
```
maxice8, line 58: can't it be a `BoolVar` instead and instead of creating a struct one just does `repoCmd.Flags().Changed("web")` ?

maxice8 (changes requested):

PR #452, opened 2020-12-30
fix(command/cmdutils): return global ID from interactive prompt

maxice8:
**Description**

MileStones need to be arranged via their global ID not their internal
one

**Related Issue**

No related issue, this is just a quick fix

Follow up for #446 

**How Has This Been Tested?**

Created issue on alpine/aports

**Types of changes**

- [ ] Documentation


profclems (approved):
Yeah I actually forgot that

PR #471, opened 2020-12-31
implement more flags for updating issues and merge requests

### LINKED ISSUE #465: allow issue update and mr update to add/remove milestones

Simple, `mr update` and `issue update` **should** be able to assign and deassign milestones

**Propose a Solution**

Not entirely clear, the one to add a milestone will obviously be `--milestone`, maybe passing an empty string or '0' to assign it to nothing ? I think we can use `cmd.Flags().Changed("milestone")` as I believe it will tell it has Changed even if the value given is nothing.

maxice8:
**Description**

- feat(commands/mr/update): add support for --milestone
- feat(commands/issue/update): add --milestone flag to assign milestones

**Related Issue**

Resolves #465

**How Has This Been Tested?**

Not thoroughly tested, just some small attempts on `gitlab.alpinelinux.org`

**Types of changes**

- [ ] Documentation

### DIFF
--- a/commands/issue/update/issue_update.go
+++ b/commands/issue/update/issue_update.go
@@ -82,6 +82,20 @@ func NewCmdUpdate(f *cmdutils.Factory) *cobra.Command {
 				actions = append(actions, "made confidential")
 				l.Confidential = gitlab.Bool(true)
 			}
+			if ok := cmd.Flags().Changed("milestone"); ok {
+				if m, _ := cmd.Flags().GetString("milestone"); m != "" || m == "0" {
+					mID, err := cmdutils.ParseMilestone(apiClient, repo, m)
+					if err != nil {
+						return err
+					}
+					actions = append(actions, fmt.Sprintf("added milestone %q", m))
+					l.MilestoneID = gitlab.Int(mID)
+				} else {
+					// Unassign the Milestone
+					actions = append(actions, "unassigned milestone")
+					l.MilestoneID = gitlab.Int(0)
+				}
+			}
 
 			fmt.Fprintf(out, "- Updating issue #%d\n", issueID)
 
@@ -107,6 +121,7 @@ func NewCmdUpdate(f *cmdutils.Factory) *cobra.Command {
 	issueUpdateCmd.Flags().StringArrayP("unlabel", "u", []string{}, "remove labels")
 	issueUpdateCmd.Flags().BoolP("public", "p", false, "Make issue public")
 	issueUpdateCmd.Flags().BoolP("confidential", "c", false, "Make issue confidential")
+	issueUpdateCmd.Flags().StringP("milestone", "m", "", "title of the milestone to assign, pass \"\" or 0 to unassign")
 
 	return issueUpdateCmd
 }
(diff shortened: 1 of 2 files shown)


commands/issue/update/issue_update.go
```diff
@@ -108,6 +159,8 @@ func NewCmdUpdate(f *cmdutils.Factory) *cobra.Command {
 	issueUpdateCmd.Flags().BoolP("public", "p", false, "Make issue public")
 	issueUpdateCmd.Flags().BoolP("confidential", "c", false, "Make issue confidential")
 	issueUpdateCmd.Flags().StringP("milestone", "m", "", "title of the milestone to assign, pass \"\" or 0 to unassign")
+	issueUpdateCmd.Flags().StringSliceP("assign", "", []string{}, "Assign new users to the issue by their `usernames`, overrides --unassign")
```
profclems: Shouldn't we be consistent with the naming of the flags? 
`issue create` uses `--assignee` so users might expect a similar name for `issue update` without reading the docs.

Also, since the `-a` shorthand isn't taken, you should give it the `-a` shorthand.

maxice8: I have been thinking whether it makes sense to have assign and unassign to add and remove assignees.

I think it is better to have assignee (replaces current assignees).

profclems: Definitely we can't have `--unassignee` 😂 so `unassign` makes sense.

Or maybe we can have a format like prefixing the username with `-` or any other sensible character to unassign a user:

```sh
glab issue update 12 --assignee "-profclems" --assignee "maxice8"

# unassigns @profclems and assigns @maxice8 
```

maxice8: How about `!` ? It is also used in `golang` to negate a boolean.

maxice8: and since we have `!` and `+` free, we can have.

--assignee `name` (removes all others and add `name`)
--assignee `+foo` (add `foo` on top of existing asssignees)
--assigne `-foo` (remove `foo` from existing assignees)

profclems: `!` is also a nice one or we can support both `-` and `!` for removing assignees and the `+` for adding to existing.

maxice8: Sounds Good, I'll drop the `assign` commit from this MR and leave only the milestone and work on it separately

profclems: I don't know if GitLab supports usernames starting with `-` (like a user has username `-anderson`) so we don't conflict with that.

profclems: I hope they don't allow that

profclems: I just checked. Only alphanumeric characters are supported

maxice8: me too, just logged out of `gitlab.com` to create a new account so I can test valid characters since I found no documentation on that. :laughing:

profclems: image

maxice8:
None of the errors from `DeepSource` are related to the code in this MR

profclems:
> None of the errors from `DeepSource` are related to the code in this MR

Yeah that's right. I'll fix those errors in a separate PR

maxice8:
Rebased

profclems (approved):

profclems:
You're free to merge when ready

maxice8:
It is ready, I am working on `--assignee` already

PR #474, opened 2020-12-31
implement --assignee with relative and absolute additions

maxice8:
**Description**
- feat(commands/cmdutils): add helper function to parse --assignee
- feat(internal/utils): add helper functions to deal with []string
- feat(commands/issue/update): add rudimentary --assignee support
- test(internal/utils): implement tests for added helper functions

**Related Issue**

Resolves #467 (Need to implement for `mr update`)

**How Has This Been Tested?**

Against `Leo/agl` on `gitlab.alpinelinux.org`

**Types of changes**

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation
- [ ] Chore (Related to CI or Packaging to platforms)

PR #479, opened 2021-01-01
Rework assignee

maxice8:
**Description**

- Rework UserAssignements as its own type
- Implement helper functions and tests for it

**Related Issue**

No related issue

Follow up to #474 

**How Has This Been Tested?**

There should be no change in behaviour aside from more accurate printing and working around GitLab's bug

**Types of changes**

- [ ] Documentation

### DIFF
--- a/commands/cmdutils/cmdutils.go
+++ b/commands/cmdutils/cmdutils.go
@@ -1,6 +1,7 @@
 package cmdutils
 
 import (
+	"errors"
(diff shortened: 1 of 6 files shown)


maxice8:
It might look like adding a lot of code but the real change is not as big as it seems plus it doesn't account the comments.

image

profclems (approved):
Nice work! Looks good

PR #482, opened 2021-01-02
Add testing for MilestonePrompt()

maxice8:
**Description**

- feat(pkg/prompt): add `Select()` to select a single option
- refactor(commands/cmdutils): use the new prompt.Select()
- test(command/cmdutils): add tests for MilestonePrompt()

**How Has This Been Tested?**

Create Merge Request with path where `cmdutils.MilestonePrompt()` is invoked


maxice8:
No clue why the tests are failing, can you lend a hand @profclems ?

maxice8:
ugh, I wrote `New release` instead of `New Release` so it was getting a nil on that map.

profclems (approved):

PR #485, opened 2021-01-03
Add more tests, use `pkg/prompt` for prompts

maxice8:
**Description**

Add tests for `BaseRepo()` and `HeadRepo()`

### DIFF
--- a/internal/git/git.go
+++ b/internal/git/git.go
@@ -477,7 +477,7 @@ func AddRemote(name, u string) (*Remote, error) {
 	}, nil
 }
 
-func SetRemoteResolution(name, resolution string) error {
+var SetRemoteResolution = func(name, resolution string) error {
 	return SetConfig(name, "glab-resolved", resolution)
 }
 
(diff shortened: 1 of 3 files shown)


profclems (approved):
LGTM! Thanks a lot

maxice8:
@profclems Is there a way to enable `group` codecov bots in this repository ?

I think with the current case only you get a `codecov-bot` comment instead of every users

profclems:
The test runs in GitLab CI but GitLab CI does not run for forks. To fix this, I did a workaround where @clemsbot creates a new branch for each PR on the GitLab mirror so the CI runs and reports back. Though the test report is sent to codecov, codecov is unable to detect the actual branch from which the PR is created from.

So to resolve this, you would have to work directly on `profclems/glab` instead of your fork

profclems:
I'm thinking of removing the test job from the GitLab CI and run it in GitHub action or Travis CI

maxice8:
I think it would be nice to have it as a GitHub action

profclems:
Yeah I will open a PR right away

PR #489, opened 2021-01-03
Resolver testing

maxice8:
**Description**

refactor(internal/glrepo/resolver.go): remove code that is always false

This code is always false because of the following code that comes a bit
before.

```golang
if r.network == nil {
    resolveNetwork(r)
    if len(r.network) == 0 {
        return nil, errors.New("no GitLab Projects found from remotes")
    }
}
```

This code will always return err if there are no `gitlab.Project` in
r.network even after using `resolveNetwork()`, and the amount of values
in `repoNames` on how many values there are in r.network.

Thus the code we are removing is never actually reached, or rather, it
is in all situations, false.

**Types of changes**

- [ ] Documentation

### DIFF
--- a/commands/cmdutils/cmdutils_test.go
+++ b/commands/cmdutils/cmdutils_test.go
@@ -600,6 +600,22 @@ func Test_PickMetadata(t *testing.T) {
 			assert.ElementsMatch(t, got, tC.expected)
 		})
 	}
+
+	t.Run("Prompt fails", func(t *testing.T) {
+		as, restoreAsk := prompt.InitAskStubber()
+		defer restoreAsk()
+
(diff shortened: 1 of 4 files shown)


maxice8:
@profclems this allows stubbed prompts to return `error` so we can test paths to handle `error`

profclems:
That's a nice one. I never thought of that

profclems (approved):

profclems:
@maxice8, can you rebase and push so we see if the test works properly on forks

profclems:
It works but codecov reported this error:
` Issue detecting commit SHA. Please run actions/checkout with fetch-depth > 1 or set to 0`.
I will open a follow up PR to #439 to fix that.

PR #492, opened 2021-01-03
No color

maxice8:


maxice8:
This makes isColorEnabled() conform to the `NO_COLOR` spec which needs the variable to be set, regardless of the value.

profclems (approved):
This is much better

PR #498, opened 2021-01-03
test(internal/glinstance/host): test HostnameValidator()

maxice8:


internal/glinstance/host_test.go
```diff
@@ -242,3 +242,55 @@ func Test(t *testing.T) {
...
+			name:     "OfficialInstance/https",
+			protocol: "https",
+			hostname: "gitlab.com",
+			output:   "https://gitlab.com/api/graphql/",
+		},
+		{
+			name:     "OfficialInstance/http",
+			protocol: "http",
+			hostname: "gitlab.com",
+			output:   "http://gitlab.com/api/graphql/",
```
profclems: with the official GitLab.com instance, the function always returns an endpoint with `https` protocol.

Here's the function:
```go
func GraphQLEndpoint(hostname, protocol string) string {
	if protocol == "" {
		protocol = "https"
	}
	if IsSelfHosted(hostname) {
		return fmt.Sprintf("%s://%s/api/graphql/", protocol, hostname)
	}
	return "https://gitlab.com/api/graphql/"
}
```

internal/glinstance/host_test.go
```diff
@@ -242,3 +242,55 @@ func Test(t *testing.T) {
...
+			name:     "OfficialInstance/https",
+			protocol: "https",
+			hostname: "gitlab.com",
+			output:   "https://gitlab.com/api/graphql/",
+		},
+		{
+			name:     "OfficialInstance/any-protocol-is-https",
+			protocol: "NoExistProtocol",
+			hostname: "gitlab.com",
+			output:   "http://gitlab.com/api/graphql/",
```
profclems:
```suggestion
			output:   "https://gitlab.com/api/graphql/",
```

maxice8:
Yes I fixed something else but forgot to fix that

PR #507, opened 2021-01-04
refactor(commands/completion): always default to bash

maxice8:


profclems (commented):
Thanks for taking this on!

It is right we default to bash if no shell is specified. We should add the to the `--shell` flag usage info that default is `bash`. Since cobra already prints the default value if set, we should set the default value to `bash`:

```go
	completionCmd.Flags().StringVarP(&shellType, "shell", "s", "bash", "Shell type: {bash|zsh|fish|powershell}")
```

maxice8:
Thanks for looking at it, I set `bash` as default outright and removed all the code that deals with the value being empty

PR #520, opened 2021-01-06
more friendly assignees experience

### LINKED ISSUE #468: make assignment on mr create more friendly

Make assignment of users on `mr create` more friendly by acquiring a List of all users that can merge to the repository and then presenting it as a multiprompt.

**Propose a Solution**

See above.

### LINKED ISSUE #472: Friendlier recovery when issue/mr create fails
When `glab mr create` fails or cancelled, the body content typed using the editor is lost.

Sometimes the user may type a long MR description and just lose it due to a simple mistake, such as not adding any commit to the MR branch yet (e.g., when creating draft MRs).

**Propose a Solution**
To me, the best user experience would be if the person can just resume where they left off so whenever someone authors an MR or issue, we save their title & body in a file scoped to that repository. 

If the create operation succeeded, we delete the file. Otherwise, the save file stays on disk and is used to pre-fill the next `issue/mr` create invocation, which is most likely a retry after a server or validation failure.

Or instead of prefilling, we provide `--recover` flag to continue from the last session

**Additional context**

maxice8:
**Description**

Create a multi-select for the user to pick assignees

**Related Issue**

Resolves #468 

**How Has This Been Tested?**

Testsuite and creating merge requests on `gitlab.alpinelinux.org`

image

**Types of changes**

- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation

This is a breaking chance because it changes how the interactive experience works. I am hesitant to call it breaking since it is interactive, not some scripting API.

### DIFF
--- a/commands/cmdutils/cmdutils.go
+++ b/commands/cmdutils/cmdutils.go
@@ -209,18 +209,55 @@ func MilestonesPrompt(response *int, apiClient *gitlab.Client, repoRemote *glrep
 	return nil
 }
 
-func AssigneesPrompt(response *[]string) (err error) {
-	var responseString string
-	err = prompt.AskQuestionWithInput(&responseString, "assignee", "Assignee(s) Username(s) [Comma Separated]", "", false)
+// GroupMemberLevel maps a number representing the access level to a string shown to the
+// user.
+// API docs:
+// https://docs.gitlab.com/ce/api/members.html#valid-access-levels
+var GroupMemberLevel = map[int]string{
+	0:  "no access",
+	5:  "minimal access",
+	10: "guest",
+	20: "reporter",
+	30: "developer",
+	40: "maintainer",
+	50: "owner",
+}
+
+// AssigneesPrompt creates a multi-selection prompt of all the users below the given access level
+// for the remote referenced by the `*glrepo.Remote`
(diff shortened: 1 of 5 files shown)


maxice8:
Another good thing is that we also respect the flags, if you pass `mr create -a profclems` and then add `maxice8` via the interactive prompt it will add both

profclems:
> Another good thing is that we also respect the flags, if you pass `mr create -a profclems` and then add `maxice8` via the interactive prompt it will add both

I've been going through the commits and really excited about this feature

profclems (commented):
We need to test this on a repo with OWNER (personal) namespace.

GitLab reports this issue under the valid access levels:

> Due to an issue, projects in personal namespaces don’t show owner (50) permission for owner.


So it is likely the owner will not be listed as part of the assignees

profclems:
> We need to test this on a repo with OWNER (personal) namespace.
> 
> GitLab reports this issue under the valid access levels:
> 
> > Due to an issue, projects in personal namespaces don’t show owner (50) permission for owner.
> 
> So it is likely the owner will not be listed as part of the assignees

Okay it works. The owner of the personal namespaced project is returned as a maintainer.

profclems (approved):
Thanks a lot for taking this on!

PR #555, opened 2021-01-10
feat(commands/variable): add `variable set` command

### LINKED ISSUE #515: Support creating project variables

It will be nice if `glab` provides support for creating project variables. Currently, users have to switch over to the GitLab web UI to create project variables

**Propose a Solution**
Provide `variable create` command for creating project variables. 
Should have the format `glab variable create <key> <value>  [flag]`. 

`new` and `set` can be used as aliases for create so `glab variable set <key> <value>  [flag]` and `glab variable set <key> <value>  [flag]` will be accepted.

Below are the parameters required by the [endpoint](https://docs.gitlab.com/ee/api/project_level_variables.html#create-variable) and the suggested flags for `glab`

Attribute | Suggested Flag | Type | required | Description
-- | -- | -- | -- | --
key | none (must be passed as first argument) |string | yes | The key of a variable; must have no more than 255 characters; only A-Z, a-z, 0-9, and _ are allowed
value | none (must be passed as second argument)  | string | yes | The value of a variable
variable_type | `-t \| --type` |string | no | The type of a variable. Available types are: env_var (default) and file
protected | `-p \| --protected` | boolean | no | Whether the variable is protected
masked | `-m \| --masked` | boolean | no | Whether the variable is masked
environment_scope | `-s \| --scope` | string | no | The environment_scope of the variable

**Additional context**

https://docs.gitlab.com/ee/api/project_level_variables.html#create-variable

profclems:
Adds `variable set` command for creating GitLab environment variables.

Variable can be created for a project or a group specified with `--group | -g` flag.

Resolves #515

### DIFF
--- a/commands/variable/set/set.go
+++ b/commands/variable/set/set.go
@@ -28,9 +28,11 @@ type SetOpts struct {
 	Protected bool
 	Masked    bool
 	Group     string
+
+	ValueSet bool
 }
 
-func NewVariableCmd(f *cmdutils.Factory, runE func(opts *SetOpts) error) *cobra.Command {
+func NewCmdSet(f *cmdutils.Factory, runE func(opts *SetOpts) error) *cobra.Command {
 	opts := &SetOpts{
 		IO: f.IO,
 	}
@@ -51,7 +53,7 @@ func NewVariableCmd(f *cmdutils.Factory, runE func(opts *SetOpts) error) *cobra.
 			$ cat file.txt | glab variable set SERVER_TOKEN
 			$ cat token.txt | glab variable set GROUP_TOKEN -g mygroup --scope=prod
 		`),
-		RunE: func(cmd *cobra.Command, args []string) error {
+		RunE: func(cmd *cobra.Command, args []string) (err error) {
 			// Supports repo override
 			opts.HTTPClient = f.HttpClient
 			opts.BaseRepo = f.BaseRepo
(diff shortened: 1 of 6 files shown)


maxice8:
small typo in the commit title, `varable->variable`

commands/variable/set/set.go
```diff
@@ -0,0 +1,170 @@
...
+		Aliases: []string{"new", "create"},
+		Args:    cobra.RangeArgs(1, 2),
+		Example: heredoc.Doc(`
+			$ glab variable set WITH_ARG "some value"
+			$ glab variable set FROM_FLAG -v "some value"
+			$ glab variable set FROM_ENV_WITH_ARG "${ENV_VAR}"
+			$ glab variable set FROM_ENV_WITH_FLAG -v"${ENV_VAR}"
+			$ glab variable set FROM_FILE < secret.txt
+			$ cat file.txt | glab variable set SERVER_TOKEN
+			$ cat token.txt | glab variable set GROUP_TOKEN -g mygroup --scope=prod
```
maxice8, line 54: Is it possible to use redirection `<` ?

something like this: `glab variable set SERVER_TOKEN < file.txt`

profclems: Yes I added an example. Check the examples

profclems:
> small typo in the commit title, `varable->variable`

Oops thanks for noticing

PR #559, opened 2021-01-11
feat(ci/view): Change color of failed pipelines that allow failure

### LINKED ISSUE #373: Pipeline views: Display failed pipelines that allow failure diferently
In commands like `pipeline status` and `pipeline view` there is no indications  that pipeline is allowed to fail e.g.
image
while in Web UI it shows up like 
image

**Propose a Solution**
Display pipelines that are allowed to fail differently, indicating that, it's okay to merge with that job failing.
For `pipeline ci view` I would propose just changing the color of said job.
In `pipeline status` it would probably be differen't text in differen't color too I supposes. Something like `passed with warnings`

**Additional context**
This would be only changes to the output of the commands, since the attribute `allow_failure` is already in the API we are using https://docs.gitlab.com/ce/api/jobs.html#list-pipeline-jobs

zemzale:
**Description**
- feat(ci/view): Display orange color on allow failure
- feat(ci/status): Change color for piplines that allow failure

**Related Issue**
Resolves #373

**How Has This Been Tested?**
Before: 
image
image

After:
image
image

**Types of changes**

- [ ] Documentation

### DIFF
--- a/commands/ci/status/status.go
+++ b/commands/ci/status/status.go
@@ -82,7 +82,11 @@ func NewCmdStatus(f *cmdutils.Factory) *cobra.Command {
 						var status string
 						switch s := job.Status; s {
 						case "failed":
-							status = utils.Red(s)
+							if job.AllowFailure {
+								status = utils.Yellow(s)
+							} else {
+								status = utils.Red(s)
+							}
 						case "success":
 							status = utils.Green(s)
 						default:
--- a/commands/ci/view/view.go
+++ b/commands/ci/view/view.go
@@ -477,8 +477,13 @@ func jobsView(app *tview.Application, jobsCh chan []*gitlab.Job, inputCh chan st
 			b.SetBorderColor(tcell.ColorGreen)
 			statChar = '✔'
 		case "failed":
-			b.SetBorderColor(tcell.ColorRed)
-			statChar = '✘'
+			if j.AllowFailure {
+				b.SetBorderColor(tcell.ColorOrange)
+				statChar = '!'
+			} else {
+				b.SetBorderColor(tcell.ColorRed)
+				statChar = '✘'
+			}
 		case "running":
 			b.SetBorderColor(tcell.ColorBlue)
 			statChar = '●'


commands/ci/status/status.go
```diff
@@ -82,6 +82,10 @@ func NewCmdStatus(f *cmdutils.Factory) *cobra.Command {
 						var status string
 						switch s := job.Status; s {
 						case "failed":
+							if job.AllowFailure {
+								status = utils.Yellow(s)
+								break
+							}
 							status = utils.Red(s)
```
profclems: Having a conditional break inside a switch which is inside a nested loop looks quite ambiguous IMO.

I would go with this
```suggestion
							if job.AllowFailure {
								status = utils.Yellow(s)
							} else {
								status = utils.Red(s)
							}
```

profclems (changes requested):
Thanks for taking this on!

Also, I'm thinking that with the `ci status`, we could add info that shows that the pipeline is allowed to fail

```sh
(success) • 1m     test                             code_navigation
(failed) • 1m      test  - allowed to fail          gemnasium-dependency_scanning
```

commands/ci/status/status.go
zemzale: Yeah, that is more readable. Changed this

zemzale:
> Also, I'm thinking that with the ci status, we could add info that shows that the pipeline is allowed to fail

Yeah that makes sense. Would be way easier to understand, but I am gonna have to then add the tabwriter there, since it's just being done with simple `fmt.Printf("\t")` and is quite messy that way. Will try to get to that in next few days

profclems:
> > Also, I'm thinking that with the ci status, we could add info that shows that the pipeline is allowed to fail
> 
> Yeah that makes sense. Would be way easier to understand, but I am gonna have to then add the tabwriter there, since it's just being done with simple `fmt.Printf("\t")` and is quite messy that way. Will try to get to that in next few days

This PR is much specific to color changes so maybe you can work on that in a separate PR and make reference to this PR

profclems (approved):

zemzale:
Sure thing

PR #560, opened 2021-01-11
chore(scripts): make quick installer script POSIX sh compliant

profclems:
Currently the quick install script isn't POSIX sh compliant and hence, works only
on bash.

In order to make it run on all shell types, it must be POSIX sh compliant.
All [bashisms][0] have been checks and removed/fixed to POSIX sh formats.

[0]: http://mywiki.wooledge.org/Bashism


scripts/quick_install.sh
```diff
@@ -1,32 +1,31 @@
-#!/usr/bin/env bash
+#!/usr/bin/env sh
```
maxice8: we can use /bin/sh as it is defined as an available path in posix

scripts/quick_install.sh
```diff
@@ -1,32 +1,31 @@
...
 
-if ! _can_install && [[ $EUID != 0 ]]; then
+if ! _can_install && [ "$(id -u)" != 0 ]; then
   echo "Run script as sudo"
   exit 1
 fi
 
 if ! _can_install; then
-  echo "Can't install to $BINDIR"
+  echo "Can't install to ${BINDIR}"
```
maxice8: `echo` isn't standard but `printf` is

```sh
    printf -- "Can't install to %s" "$BINDIR"
```

scripts/quick_install.sh
profclems: Okay sure I will change that

scripts/quick_install.sh
profclems: Thanks. Fixed that

PR #578, opened 2021-01-18
chore(deps): Update rivo/tview to newest verison

zemzale:
**Description**
Updates `rivo/tview` to the newest version. 

**Related Issue**
Resolves #561


profclems (approved):
Thanks for following up on the issue!

I think we may have to close #561 since this PR updates both `tview` and `runewidth` packages

zemzale:
> I think we may have to close #561 since this PR updates both tview and runewidth packages

Yeah, that was the plan after test had passed, but You already did that.

PR #584, opened 2021-01-20
feat(commands/mr/create): use tracking branch as target branch if not given

### LINKED ISSUE #582: mr create doesn't respect upstream branch as target branch
**Description**

When using `mr create` without `--target-branch` it will always create against the main branch of the remote.

It should respect the upstream branch and use it if defined.

**Possible Fix**

make `mr create` check if there is a upstream remote/branch defined and set the upstream branch as the target branch

**Your Environment**

* Version used (Run `glab --version`): glab version 1.13.1-375-g6acef3d (2021-01-18)
* Operating System and version: Fedora Silverblue 33

maxice8:
**Description**

If --target-branch is not given and there is tracking branch, use it.

**Related Issue**

Resolves #582 

**How Has This Been Tested?**

Tested against `gitlab.alpinelinux.org`

**Types of changes**

- [ ] Documentation

### DIFF
--- a/commands/mr/create/mr_create.go
+++ b/commands/mr/create/mr_create.go
@@ -205,20 +205,7 @@ func NewCmdCreate(f *cmdutils.Factory) *cobra.Command {
 			}
 
 			if opts.TargetBranch == "" {
-				branchConfig := git.ReadBranchConfig(opts.SourceBranch)
-				// Check if our given git.BranchConfig{} is not empty, otherwise it will fail
-				// if try to access the fields, this is needed because the ReadBranchConfig
-				// function can return an empty struct
-				if branchConfig != (git.BranchConfig{}) {
-					if branchConfig.RemoteName != "" && branchConfig.MergeRef != "" {
-						// The MergeRef takes the form of refs/head/BRANCH_NAME, so split it
-						// by '/' and get the last element
-						branchName := strings.Split(branchConfig.MergeRef, "/")
-						opts.TargetBranch = branchName[len(branchName)-1]
-					}
-				} else {
-					opts.TargetBranch, _ = git.GetDefaultBranch(baseRepoRemote.PushURL.String())
-				}
+				opts.TargetBranch = getTargetBranch(baseRepoRemote, opts.SourceBranch)
 			}
 
 			opts.TargetTrackingBranch = fmt.Sprintf("%s/%s", baseRepoRemote.Name, opts.TargetBranch)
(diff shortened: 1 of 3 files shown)


commands/mr/create/mr_create.go
```diff
@@ -209,6 +204,20 @@ func NewCmdCreate(f *cmdutils.Factory) *cobra.Command {
 				}
 			}
 
+			if opts.TargetBranch == "" {
+				// Check if have a tracking branch (MergeRef) and use it as target branch
+				branchConfig := git.ReadBranchConfig(opts.SourceBranch)
+				if branchConfig.RemoteName != "" && branchConfig.MergeRef != "" {
```
profclems: `git.ReadBranchConfig` is likely to return a `nil` if an error occurs. See

https://github.com/profclems/glab/blob/b2922a72dcf07f25a95046cdd3e9ff0afc2e1ff3/internal/git/git.go#L234-L240

 so it is important to check if `branchConfig` is not nil before accessing the fields to avoid panic

profclems (changes requested):
Thanks for working on this.

I have a little issue here:

commands/mr/create/mr_create.go
maxice8: I tested if it returns something that isn't `(git.BranchConfig{})`, hope that is enough

commands/mr/create/mr_create.go
```diff
@@ -209,6 +204,24 @@ func NewCmdCreate(f *cmdutils.Factory) *cobra.Command {
...
+				if branchConfig != (git.BranchConfig{}) {
+					if branchConfig.RemoteName != "" && branchConfig.MergeRef != "" {
+						// The MergeRef takes the form of refs/head/BRANCH_NAME, so split it
+						// by '/' and get the last element
+						branchName := strings.Split(branchConfig.MergeRef, "/")
+						opts.TargetBranch = branchName[len(branchName)-1]
+					}
+				} else {
+					opts.TargetBranch, _ = git.GetDefaultBranch(baseRepoRemote.PushURL.String())
+				}
```
profclems, line 221:
```suggestion
				if branchConfig != nil {
					if branchConfig.RemoteName != "" && branchConfig.MergeRef != "" {
						// The MergeRef takes the form of refs/head/BRANCH_NAME, so split it
						// by '/' and get the last element
						branchName := strings.Split(branchConfig.MergeRef, "/")
						opts.TargetBranch = branchName[len(branchName)-1]
					} else {
						opts.TargetBranch, _ = git.GetDefaultBranch(baseRepoRemote.PushURL.String())
					}
				} else {
					opts.TargetBranch, _ = git.GetDefaultBranch(baseRepoRemote.PushURL.String())
				}
```

profclems (approved):

PR #598, opened 2021-01-30
feat: adds possibility to override host for API

### LINKED ISSUE #350: Cannot get glab working on private instance with different endpoints
We are running our own gitlab instance and the api/http have a different address then the ssh

git.name.devops.com  (ssh)
gitlab.name.devops.com (webUi and api)

**Description**
Using glab auth login I can login but I am not able to use any of the functionality like list issues

**Expected Behavior**

Be able to use the great features
**Actual Behavior**

glab says `none of the git remotes configured for this repository points to a known GitLab host. Please use `glab auth login` to authenticate and configure a new host for glab`
**Possible Fix**

**Steps to Reproduce**

1. glab auth login
-- u are already logged in
2. glab issue list
none of the git remotes configured for this repository points to a known GitLab host. Please use `glab auth login` to authenticate and configure a new host for glab

**Context**

Can't use the glab cli
**Your Environment**

* Version used:  glab version 1.12.1
Ubuntu 18.04 &  zsh

lkwg82:
**Description**
This PR adds the possibility to add an `api_host` entry in the config to enable overriding the default repository host.

In my company we split the api frontend from the ssh backend in the dns. So the api tries to call the ssh git backend.

**Related Issue**

I'm not sure, if I missed one.

**How Has This Been Tested?**
I debugged and started with non-working to working code.

**Types of changes**

- [ ] Documentation


profclems (commented):
Thanks for the contribution! In the future, it would be very useful to link to an issue that this is solving. We generally do not accept PRs that are not linked to any particular issue.

Would this be addressing #350?

profclems (commented):
This seems to be solving #350

I'm really excited about this. The current changes look good. You can proceed to add the necessary documentation.

The `api_host` becomes directly attached to the authenticated host
```yaml
hosts:
    git.name.devops.com:
          api_host: gitlab.name.devops.com
          git_protocol: ssh
          token: xxxxx
...
```

Note that `cfg.Get(repoHost, "api_host")` gets the value in this order:
```
ENV VARS -> config(hosts) -> config(global or non-host) 
```

profclems:
Also, we may have to add `api_host` to our default config mapping for the `hosts`:
https://github.com/profclems/glab/blob/222c0c51dd4288189fba8f50b1dd6118029ffe22/internal/config/config_type.go#L215-L244

lkwg82:
@profclems I need help with the tests, I dont understand the relation.

internal/config/config_type.go
```diff
@@ -226,6 +226,11 @@ func NewBlankRoot() *yaml.Node {
 										Kind:        yaml.ScalarNode,
 										Value:       "api_protocol",
 									},
+									{
+										HeadComment: "Configure host for api endpoint, defaults to the host itself",
+										Kind:        yaml.ScalarNode,
+										Value:       "api_host",
+									},
```
profclems, line 237: With this, you've only set the key without the value so it reads as `api_protocol: api_host` causing most tests to fail.
Here is a suggestion:

```suggestion
									{
										HeadComment: "Configure host for api endpoint, defaults to the host itself",
										Kind:        yaml.ScalarNode,
										Value:       "api_host",
									},
									{
										Kind:  yaml.ScalarNode,
										Value: "gitlab.com",
									},
```
It should also be above the api_protocol (i.e before line 224)

profclems (changes requested):

lkwg82:
Afaik everything is resolved, did I miss sth?

README.md
```diff
@@ -315,6 +315,8 @@ $ glab config set editor vim --host gitlab.example.org
 
   GITLAB_URI or GITLAB_HOST: specify the url of the gitlab server if self hosted (eg: https://gitlab.example.com). Default is https://gitlab.com.
 
+  GITLAB_API_HOST: specify the host where the endpoint is found: defaults to the hostname
```
profclems:
```suggestion
  GITLAB_API_HOST: specify the host where the API endpoint is found. Useful when there are separate [sub]domains or hosts for git and the API endpoint: defaults to the hostname found in the git URL
```

Same thing should be at 
https://github.com/profclems/glab/blob/222c0c51dd4288189fba8f50b1dd6118029ffe22/commands/root.go#L37-L60

README.md
```diff
@@ -331,6 +333,25 @@ $ glab config set editor vim --host gitlab.example.org
...
+as config in global config file
+```yaml
+...
+hosts:
+    mygitlab.com:
+        api_host: api.mygitlab.com
+        token: mytoken
+...
+```
+
```
profclems: Considering that we haven't added examples for the other env vars, the examples here won't be needed as it should be obvious to users on what to do.
The `api_host` can be set with `glab config set api_host api.mygitlab.com -h mygitlab.com`
```suggestion
```

I plan of modifying `glab auth login` so that user is prompted to set the `api_host`

profclems (changes requested):
Almost there @lkwg82!

Just a few things to improve upon

profclems:
I will open a follow-up PR with changes to `glab auth login` so that the user is prompted to set the `api_host`

README.md
lkwg82: I added an example, because it was not that obvious. 

Even the command `glab config set api_host api.mygitlab.com -h mygitlab.com` which adds a host in the global config is not obvious besides the '--global' flag for configs. ;)

README.md
lkwg82: why is the documentation duplicated and slightly different?

in README
```
GITLAB_URI or GITLAB_HOST: specify the url of the gitlab server if self hosted (eg: https://gitlab.example.com). Default is https://gitlab.com.
```

in root.go
```
GITLAB_HOST or GL_HOST: specify the url of the gitlab server if self hosted (eg: https://gitlab.example.com). Default is https://gitlab.com.
```

lkwg82: I assume some ENV Variables are kind of `old` and should be deprecated to clean out the logic and documentation.
What about introducing a way to print out a warning when the usage of old vars is detected? (in a separate PR)

lkwg82: Can we generate one out the other? Or just point to the other documentation? Or include it, I know in ASCII Doc this is possible. The current state feels not right.

README.md
lkwg82: While reading through the code I see some opportunities for refactorings:
- centralise config (I see glinstance.* invocation very deep down the invocation chain)
- maybe introduce a logger to enable debug logs overall the place

just my two cents

README.md
profclems: Sorry I couldn't get back to you earlier. Had to take some break since it's Sunday

profclems: I know the current state is a bit off. I had actually thought about refactoring but the readme uses markdown and it's not possible to centralise it. Maybe we would have to remove it from the docs so the users rely on the `help` command (But what about those using it in their CI environment only)

profclems: > What about introducing a way to print out a warning when the usage of old vars is detected? (in a separate PR)

I agree with this... Should be a separate PR

README.md
profclems: > I added an example, because it was not that obvious.

Should be obvious as long as the `GITLAB_API_HOST` is indicated as part of the list of respected env vars and it is added to the config file well.

From my comment here: https://github.com/profclems/glab/pull/598#issuecomment-770300435
> I will open a follow-up PR with changes to glab auth login so that the user is prompted to set the api_host

`glab auth login` will have `--api-host` flag as well

profclems: > - maybe introduce a logger to enable debug logs overall the place

Hmm... You're right on this! We can delve deeper into this in a separate issue/PR

.github/CONTRIBUTING.md
```diff
@@ -33,6 +33,8 @@ Run the new binary as: `./bin/glab`
 
 Run tests with: make test or go test ./...
 
+WARNING: Do not run `make test` outside of an isolated environment, it will overwrite your global config.
```
profclems: I would that you drop this commit and open it in a separate PR... I prefer having PRs focused on its description
```suggestion
```

profclems (commented):
Thanks a lot @lkwg82... I really appreciate your effort to help make glab better

One last thing before I merge this PR

profclems (approved):
Thanks for taking in this one @lkwg82

PR #601, opened 2021-02-01
Add prompt to git init on repo create. Fixes #569

### LINKED ISSUE #569: Prompt to 'git init' on repo create

If you run `glab repo create` in a directory that isn't initialized with git the command fails:
```sh
→ glab repo create
fatal: not a git repository (or any of the parent directories): .git
git: exit status 128
```

It is required that the user first run `git init` before running `glab repo create`. 

**Propose a Solution**

If the directory isn't git initialized, the user should be prompted whether `glab` should run `git init` before creating the repo
```sh
? Directory not git initialized. Run `git init`? [Y/N]
```

**Additional context**

cvhariharan:
**Description**
Adds a prompt to run `git init` if the directory is not initialized before running `glab repo create`.

**Related Issue**
Resolves #569 

**How Has This Been Tested?**
- Created an empty directory without running `git init` and running `glab repo create` directly. -> Prompt for git init shown
- Created an empty directory and ran `git init` before running `glab repo create`. -> No prompt shown

**Types of changes**

- [ ] Documentation

### DIFF
--- a/commands/project/create/project_create.go
+++ b/commands/project/create/project_create.go
@@ -210,7 +210,7 @@ func initGit() error {
 		return nil
 	}
 	var doInit bool
-	err := prompt.Confirm(&doInit, fmt.Sprintf("Directory not git initialized. Run `git init`?"), true)
+	err := prompt.Confirm(&doInit, "Directory not git initialized. Run `git init`?", true)
 	if err != nil || !doInit {
 		return err
 	}
--- a/commands/project/create/project_create.go
+++ b/commands/project/create/project_create.go
@@ -68,7 +68,11 @@ func runCreateProject(cmd *cobra.Command, args []string, f *cmdutils.Factory) er
 	)
 	c := f.IO.Color()
 
-	err = initGit()
+	defaultBranch, err := cmd.Flags().GetString("defaultBranch")
+	if err != nil {
+		return err
+	}
+	err = initGit(defaultBranch)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


commands/project/create/project_create.go
```diff
@@ -199,6 +205,22 @@ func runCreateProject(cmd *cobra.Command, args []string, f *cmdutils.Factory) er
...
+	if stat, err := os.Stat(".git"); err == nil && stat.IsDir() {
+		return nil
+	}
+	var doInit bool
+	err := prompt.Confirm(&doInit, fmt.Sprintf("Directory not git initialized. Run `git init`?"), true)
+	if err != nil || !doInit {
+		return err
+	}
+
+	gitInit := git.GitCommand("init")
```
profclems, line 221: Don't you think we should consider `git init`-ializing with the default branch provided by the user via the `--defaultBranch` flag?

If so, the `initGit` function signature will change to `initGit(branch string)`.

```go
gitInit := git.GitCommand("init", "-b", branch)
```

On line 71, you would have to then call the `initGit` function with the default branch provided by the user
```go
err = initGit(cmd.Flags().GetString("defaultBranch"))
if err != nil {
	return err
}
```

profclems (changes requested):
Thanks a lot @cvhariharan for your contribution. Sorry I couldn't review much earlier.

I have just one issue that should be fixed before merging this PR

commands/project/create/project_create.go
cvhariharan: Yeah I totally agree. Didn't notice that a default branch could be passed. I'll make the change.

profclems: Great!

cvhariharan: @profclems Instead of using the `-b` flag I have added a checkout command after git init if the default branch is not empty. This is to ensure backwards compatibility with older versions of git.

profclems: Makes perfect sense! Thanks

profclems (approved):
Thanks @cvhariharan  for taking this on... LGTM

cvhariharan:
Thanks @profclems. This is a really cool and useful project. Happy to help :smile:.

PR #605, opened 2021-02-03
fix(issue): Open issue web view instantly if, URL provided

### LINKED ISSUE #556: glab issue view --web <URL> is not instant
**Description**

When using --web and providing a <URL> instead of just a number there shouldn't be any delay between it and calling the browser to open.

It takes up to a full second to open it in the browser

**Possible Fix**

Check if given ARG is an URL and skip over any and all network requests or other measures that might be wasting time

**Logs**

```console
foo@bar:~$ time glab issue view --web https://gitlab.alpinelinux.org/alpine/aports/-/issues/1
Opening gitlab.alpinelinux.org/alpine/aports/-/issues/1 in your browser.

________________________________________________________
Executed in    1.39 secs   fish           external
   usr time   79.19 millis  349.00 micros   78.85 millis
   sys time   25.93 millis   62.00 micros   25.86 millis
```

**Your Environment**

* Version used (Run `glab --version`): glab version 1.13.1-335-g73abc63 (2021-01-10)
* Operating System and version: Fedora Silverblue 33

zemzale:
**Description**
When opening issue from URL, we should skip all the API calls, and just instantly open up.

**Related Issue**

Resolves #556

Note: Issue mentions the same thing for `mr` command, but that command simply doesn't open URLs to begin with. That should be implemented on it's own issue.

**How Has This Been Tested?**

Before :
```sh
glab issue view --web https://gitlab.com/smth/backend/-/issues/2  0.13s user 0.05s system 18% cpu 0.949 total
```
Now:
```sh
glab issue view --web https://gitlab.com/smth/backend/-/issues/2  0.07s user 0.03s system 68% cpu 0.146 total
```

**Types of changes**

- [ ] Documentation

### DIFF
--- a/commands/issue/view/issue_view.go
+++ b/commands/issue/view/issue_view.go
@@ -51,12 +51,18 @@ func NewCmdView(f *cmdutils.Factory) *cobra.Command {
 		`),
 		Args: cobra.ExactArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
+			cfg, _ := f.Config()
+			_, baseRepo := issueutils.IssueMetadataFromURL(args[0])
+
+			if opts.Web && baseRepo != nil {
+				browser, _ := cfg.Get(baseRepo.RepoHost(), "browser")
+				return openWeb(args[0], browser, f.IO)
+			}
 
 			apiClient, err := f.HttpClient()
 			if err != nil {
 				return err
 			}
-			cfg, _ := f.Config()
 
 			issue, baseRepo, err := issueutils.IssueFromArg(apiClient, f.BaseRepo, args[0])
 			if err != nil {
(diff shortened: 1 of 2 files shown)


commands/issue/view/issue_view.go
```diff
@@ -114,6 +116,13 @@ func NewCmdView(f *cmdutils.Factory) *cobra.Command {
 	return issueViewCmd
 }
 
+func openWeb(url string, browser string, io *iostreams.IOStreams) error {
```
profclems, line 119: I'd prefer you make `io *iostreams.IOStream` the first parameter
```suggestion
func openWeb(io *iostreams.IOStreams, url string, browser string) error {
```

profclems (changes requested):
Thanks for working on this

maxice8:
What if we give an URL to an issue that does not exist?

profclems:
> What if we give an URL to an issue that does not exist?

I'm thinking about this too!

This will still definitely require making network requests rendering #556 invalid

profclems:
https://github.com/profclems/glab/blob/d5179624227f06f44c98cf14518414dfc92ff5cf/commands/issue/issueutils/utils.go#L138 only uses regex to check if the URL provided matches the GitLab merge request link format.
It makes no network request to check the validity

commands/issue/view/issue_view.go
zemzale: Sure thing. I will fix this first thing tomorrow morning.

zemzale:
If we wanna check if the URL we provided is an existing, and valid issue, that would just be the current thing we do.

profclems:
> If we wanna check if the URL we provided is an existing, and valid issue, that would just be the current thing we do.

Yeah just as I said.
Which will render the issue we're trying to resolve (#556) invalid

maxice8:
Yes, if we are giving the full URL why even bother using `glab`, just do `open <URL>`

zemzale:
I am up for just not merging the commit, to keep the code simpler. We really can open up only full URLs, and then it doesn't make sense to use `glab` as fancy `open`

maxice8:
yes

zemzale:
Then I am just gonna close this PR, as not acceptable

PR #627, opened 2021-02-24
feat(project.clone): add option to preserve namespaces by subdirectories

loupgaroublond:
**Description**

**Related Issue**

Resolves #[issue_number]

**How Has This Been Tested?**

**Types of changes**

- [ ] Documentation

### DIFF
--- a/commands/project/clone/repo_clone.go
+++ b/commands/project/clone/repo_clone.go
@@ -22,6 +22,7 @@ import (
 type CloneOptions struct {
 	GroupName         string
 	IncludeSubgroups  bool
+	PreserveNamespace bool
 	WithMREnabled     bool
 	WithIssuesEnabled bool
 	WithShared        bool
@@ -88,7 +89,7 @@ Clone a GitLab repository/project
 		RunE: func(cmd *cobra.Command, args []string) error {
 			if nArgs := len(args); nArgs > 0 {
 				ctxOpts.Repo = args[0]
-				if nArgs > 1 {
+				if nArgs > 1 && !opts.PreserveNamespace {
 					opts.Dir = args[1]
 				}
 				opts.GitFlags = args[1:]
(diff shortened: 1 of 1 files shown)


zemzale:
Thanks, for the PR @loupgaroublond 
We generally don't accept PRs without an issue attached to them, but this seems, like a nice feature, and does work as expected.

If @profclems is up for this I believe this can be merged into trunk

loupgaroublond:
If you folks can take this as is, that is excellent. Otherwise, happy to open up an issue and go through the change more. There were definitely a few behavioral choices I made unilaterally, that, both, I'd like to change, and I bet others have Strong Opinions on too.

If nothing else, we needed this change at work - so here's the change, it'll get active use off my fork, and whenever you're ready, we can merge it back in.

profclems:
Thanks for working on this @loupgaroublond 

@zemzale you may review and merge if you have some time on your hands right now

profclems (commented):
@loupgaroublond thanks for working on this! It's a really nice feature.
I'm having a few thoughts on the naming of the `--preserve-namespace | p` flag. 

@zemzale @maxice8 should we maintain that flag name?

zemzale (commented):
I don't have a better idea for `preserve-namespace`, name, but we might go without the `p` short flag maybe? I don't think that this flag is such a common thing, and there are already a lot of instances of `p` mostly meaning `page` for list requests.

zemzale:
This also works with nested groups, 
image
Turns into 
image

Not counting the nitpick about the short flag LGTM.

zemzale (approved):

loupgaroublond:
> I don't have a better idea for `preserve-namespace`, name, but we might go without the `p` short flag maybe? I don't think that this flag is such a common thing, and there are already a lot of instances of `p` mostly meaning `page` for list requests.

I am ok with dropping the short flag. Please stay tuned for a patch to the PR.

loupgaroublond:
> This also works with nested groups,
> image
> Turns into
> image
> 
> Not counting the nitpick about the short flag LGTM.

Those nested groups is entirely the motivating use case for me here :)

profclems (approved):
On second thought, we can still maintain the `-p` shorthand for the `--preserve-namespace` flag since this isn't a `list` command

Thanks a lot @loupgaroublond for working on this... It will be released as part of v1.16.0

PR #667, opened 2021-03-25
Fix hostnames resolution

### LINKED ISSUE #592: GITLAB_{URI,HOST} lead to failure if scheme is provided, documentation doesn't match what code needs
**Description**

We have a self-hosted GitLab instance.

The `glab` help string that says:

>   GITLAB_HOST or GL_HOST: specify the url of the gitlab server if self hosted (eg: https://gitlab.example.com). Default is https://gitlab.com.

If I set GITLAB_HOST to `https://gitlab.my.company.com`, it fails to connect to my GitLab:

```
▶ export GITLAB_HOST=https://gitlab.my.company.com
▶ ./bin/glab repo search -s docker
error connecting to https
lookup https on 10.0.1.11:53: read udp 172.16.194.177:48449->10.0.1.11:53: i/o timeout
check your internet connection or status.gitlab.com or 'Run sudo gitlab-ctl status' on your server if self-hosted

▶
```

Walking into the code with Delve, I can see where the API endpoint is screwy:

```
> github.com/profclems/glab/internal/glinstance.APIEndpoint() ./internal/glinstance/host.go:66 (PC: 0x6cb5e7)
    61:	func APIEndpoint(hostname, protocol string) string {
    62:		if protocol == "" {
    63:			protocol = "https"
    64:		}
    65:		if IsSelfHosted(hostname) {
=>  66:			return fmt.Sprintf("%s://%s/api/v4/", protocol, hostname)
    67:		}
    68:		return "https://gitlab.com/api/v4/"
    69:	}
    70:
    71:	// GraphQLEndpoint returns the GraphQL API endpoint prefix for a GitLab instance :)
(dlv)
> github.com/profclems/glab/pkg/api.(*Client).NewLab() ./pkg/api/client.go:214 (PC: 0xebc094)
Values returned:
	~r2: "https://https://gitlab.my.company.com/api/v4/"
```

If I set GITLAB_HOST (also tested GITLAB_URI) to *just the hostname*, then it works as expected

```
▶ export GITLAB_HOST=gitlab.my.company.com
▶ ./bin/glab repo search -s docker
Showing results for "docker"
[...]
```

**Possible Fix**

Change the documentation string to describe what is actually acceptable, or change the code to handle the string that the documentation suggests.

**Steps to Reproduce**

see above.

**Your Environment**

CentOS 7, go 1.13.11 (sigh...)

Glab built locally from source

```
▶ ./bin/glab --version
[--version] -> [--version]
glab version 1.14.0 (2021-01-28)
▶ git log | head
commit a086fdd49409da8b621f10c0cd2778b4d5b2f51a
[...]
```

zemzale:
**Description**
When overriding host, we now strip protocol from the URL, since we don't need that and it would break config key lookups. 

**Related Issue**
Resolves #592

**How Has This Been Tested?**

Locally tested with self-hosted GitLab instance

**Types of changes**

- [ ] Documentation

### DIFF
--- a/cmd/glab/main.go
+++ b/cmd/glab/main.go
@@ -22,6 +22,7 @@ import (
 	"github.com/profclems/glab/internal/glinstance"
 	"github.com/profclems/glab/internal/run"
 	"github.com/profclems/glab/pkg/tableprinter"
+	"github.com/profclems/glab/pkg/utils"
 
 	"github.com/spf13/cobra"
 )
@@ -184,6 +185,9 @@ func maybeOverrideDefaultHost(f *cmdutils.Factory) {
 		glinstance.OverrideDefault(baseRepo.RepoHost())
 	}
 	if glHostFromEnv := config.GetFromEnv("host"); glHostFromEnv != "" {
+		if utils.IsValidURL(glHostFromEnv) {
+			glHostFromEnv, _ = glinstance.StripHostProtocol(glHostFromEnv)
+		}
 		glinstance.OverrideDefault(glHostFromEnv)
 	}
 }


internal/glinstance/host.go
```diff
@@ -63,6 +63,9 @@ func APIEndpoint(hostname, protocol string) string {
 		protocol = "https"
 	}
 	if IsSelfHosted(hostname) {
+		if strings.HasPrefix(hostname, "http") {
```
profclems: This isn't a practical operation taking URLs like [httpie.io](https://httpie.io) and others with `http` prefix as part of the hostname. You can use [`utils.IsValidURL(hostname)`](https://github.com/profclems/glab/blob/6e21a963a9af9ce0ea76bc7db951e3df22db4419/pkg/utils/utils.go#L141) found in our [utils](https://github.com/profclems/glab/blob/6e21a963a9af9ce0ea76bc7db951e3df22db4419/pkg/utils/utils.go#L141) package.

profclems (changes requested):

internal/glinstance/host.go
zemzale: This was just a dirty fix I so I can display better the issues I encountred. 

I wrote up comment about this https://github.com/profclems/glab/issues/592#issuecomment-807012822 
TDLR: This would fix the problem of https://https://hostname, but config resolution would break since we are storing configs with hostnames without the protocol

zemzale: I change the approach of fixing this completely so this is not relevant anymore.

profclems (approved):

PR #684, opened 2021-04-16
Add repo list

### LINKED ISSUE #671: glab repo list

same as gh repo list, show all current active repos on the active account.

**Propose a Solution**

$ glab repo list
**Additional context**

olearycrew:
**Description**
This change adds a command `glab repo list` that will list the user's owned repos (projects) by any chosen order / sort like other lists - similar to and modeled after the `glab repo contributors` list.

**Related Issue**
Resolves #671

**How Has This Been Tested?**
I tested this with my own account, using various orders and sort directions.

image

**Types of changes**
- [ ] Documentation


jeanphi-baconnais:
Hi @olearycrew , i've check your PR and it's seems to be OK 👍

commands/project/list/list.go
```diff
@@ -0,0 +1,95 @@
...
+	c := opts.IO.Color()
+
+	apiClient, err := opts.HTTPClient()
+	if err != nil {
+		return err
+	}
+
+	l := &gitlab.ListProjectsOptions{
+		OrderBy:        gitlab.String(opts.OrderBy),
+		MinAccessLevel: gitlab.AccessLevel(50), // Only projects you own
```
zemzale, line 68: I think it would be better to set this to Minimal access instead?

I tried to run this against my GitLab.com account. That showed me most of my projects(there is a [bug](https://gitlab.com/gitlab-org/gitlab/-/issues/219299)).

Then I tried to run against one of my self hosted instances, and it showed me nothing. When I go into Web UI I have 5 projects available, and I expected to see the same. The difference comes from the fact that I am just a developer on these projects.

And I feel that a lot of people might have such instances where they are only developers on a lot of projects and would want to see them, but can't.

zemzale (changes requested):
I looked over the PR and have one slight issue with what projects are actually being shown.

Other than that this looks fantastic.

commands/project/list/list.go
profclems: Yeah it tried it too on GitLab.com and it only show repos I own without including projects in my personal namespace

```
$ glab repo list

Showing 2 of 2 projects (Page 1 of 1)

glab-cli/test     
glab-cli/test-pv  
```

I suggest instead of restricting the result with the min access level, we should have flags to allow the user get the type of repos to return
```go
repoListCmd.Flags().BoolVarP(&opts.All, "all", "a", false, "list all projects on the instance")
repoListCmd.Flags().BoolVarP(&opts.Owned, "mine", "m", false, "limit by projects you own")
repoListCmd.Flags().BoolVar(&opts.Member, "member", false, "limit by projects which you are a member")
repoListCmd.Flags().BoolVar(&opts.Starred, "starred", false, "Get starred projects")
```

ghost (commented):
Duplicate of #

ghost (commented):
Duplicate of #

profclems:
Closing in favour #906 which still has @olearycrew's commits

PR #695, opened 2021-04-29
feat: mr merge options

profclems:
This adds merge options when merging an MR. The user is given three merge options
- create merge commit
- squash and merge
- rebase and merge

NB: This options are only available on TTYs. On non-ttys, the default is to create a merge commit unless the user explicitly specifies the merge method using the available flag options

### DIFF
--- a/commands/cmdutils/cmdutils.go
+++ b/commands/cmdutils/cmdutils.go
@@ -99,7 +99,7 @@ func GetEditor(cf func() (config.Config, error)) (string, error) {
 	return editorCommand, nil
 }
 
-func DescriptionPrompt(response *string, templateContent, editorCommand string) error {
+func EditorPrompt(response *string, question, templateContent, editorCommand string) error {
 	defaultBody := *response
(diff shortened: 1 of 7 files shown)


zemzale (approved):
Looks nice!

maxice8:
Thanks for working on this but it still has the race condition problem:

```console
[container::toolbox] ───── gl mr merge garcon                                                 ~/R/aports (glib-networking)
? Merge when pipeline succeeds? No
? What merge method would you like to use? Rebase and merge
? What's next? Submit
✓ Rebase successful
\ Merging merge request !20994 PUT https://gitlab.alpinelinux.org/api/v4/projects/alpine/aports/merge_requests/20994/merge: 406 {message: Branch cannot be merged}
```

Trying to merge too quickly after rebasing results in this error, I work around this on my wrapper shellscript to just keep trying 2 or 3 times every 4 seconds

profclems:
Thanks @maxice8... I will fix that with a follow-up PR

PR #678, opened 2021-04-08
Fix: Allow for override of repo using OWNER/REPO

zemzale:
**Description**
When using flags/commands that override the repo, but don't supply
hostname, we allways fallback to `gitlab.com`. This is a problem for
commands like `mr checkout 12 -R OWNER/REPO` that should use API host
`self-hosted`, but actually use `gitlab.com`.

Now we override default host, from the base repo, if one is found.
Environment values still override repo host.

**Related Issue**
Resolves #383

**How Has This Been Tested?**
Locally tried to checkout mr using `mr checkout 12 -R ONWER/REPO`, on self hosted instance and everything worked.

**Types of changes**

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation
- [ ] Chore (Related to CI or Packaging to platforms)

PR #696, opened 2021-04-30
Allow protocol override when overriding the default host

profclems:
This allows overriding the default protocol of the hostname if a fully qualified url is provided when using the `GITLAB_HOST` environment variable.
For example, if the GITLAB_HOST=http://my.glinstance.org, then the host will be set to my.glinstance.org and the protocol become http, unlike before when the protocol was ignored and only the hostname was considered.

This is a follow-up for #678 and #667


zemzale (approved):
Nice work @profclems . 

I had the idea of protocol override in the back of my head when I worked on this, but in the end, I forgot about it.

Issue #708, opened 2021-05-11
[Refactor] Project create


zemzale:
**Describe the feature or problem you'd like to solve**

We need to refactor and add test to the `repo create` command. There are already tests, but they are commented out since I assume they were hard to run.

**Propose a Solution**

- Refactor `repo_create.go` similar to something like `mr_create.go`
- Add proper tests for the command

clemsbot:
This issue has been automatically marked as stale because it has not had recent activity. We haven't had the time to address it yet, but we want to keep it open. This message is just a reminder for us to help triage issues.

clemsbot:
This issue has been automatically marked as stale because it has not had recent activity. We haven't had the time to address it yet, but we want to keep it open. This message is just a reminder for us to help triage issues.

PR #709, opened 2021-05-11
feat(repo): Create project from URL

### LINKED ISSUE #618: the project create command currently does not support full urls as parameter
the project create command currently does not support full urls as parameter

**Propose a Solution**
Allow for it

**Additional context**
https://github.com/profclems/glab/issues/609#issuecomment-779973378

Additionally, would any of the settings in the config file allow for overwriting the default host setting?

zemzale:
**Description**
This is gonna allow for creating projects using full URL which is gonna
override the host URL too.

The command doesn't have unit tests and probably should be refactored, so I opened issue about that #708

**Related Issue**
Resolves #618

**How Has This Been Tested?**
Locally tested

**Types of changes**

- [ ] Documentation

### DIFF
--- a/commands/project/create/project_create.go
+++ b/commands/project/create/project_create.go
@@ -10,7 +10,6 @@ import (
 	"github.com/profclems/glab/pkg/prompt"
 
 	"github.com/profclems/glab/api"
-	"github.com/profclems/glab/internal/glinstance"
 	"github.com/profclems/glab/internal/glrepo"
 
 	"github.com/MakeNowJust/heredoc"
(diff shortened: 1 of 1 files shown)


profclems (changes requested):
Thanks for working on this @zemzale.

I tested this and got a few issues
- This works fine when the argument provided is a URL (https://gitlab.com/profclems/test)
- Does not support ssh urls (e.g. git@gitlab.com:profclems/test.git)
- Does not support urls with `.git` prefix (https://gitlab.com/profclems/test.git)
- It disregards the hostname and uses the default host `gitlab.com`

zemzale:
Indeed these cases wont work. I will fix these issues, shouldn't be too hard, I just didn't have the chance to test against a private instance.

profclems:
I think it should be easier with this function in the [`glrepo`](https://github.com/profclems/glab/blob/a17c5625128cdcad0dcb430df633b91ea7f379b6/internal/glrepo/repo.go) package

https://github.com/profclems/glab/blob/a17c5625128cdcad0dcb430df633b91ea7f379b6/internal/glrepo/repo.go#L117-L183

profclems:
@zemzale I just pushed a commit to fix the issues I listed

profclems (approved):

PR #729, opened 2021-05-16
fix job viewing with detached pipelines

### LINKED ISSUE #721: Support for detached pipeline in `ci trace`

When running a `glab ci trace`, I get the error message `No pipeline running or available on <branch>`. However, there is a detached pipeline on that branch, or rather on the MR of that branch.

Interestingly, `glab ci view` works fine, so I guess the logic for matching that already exists?
Or maybe I'm missing something and am using it wrong :-) 

**Propose a Solution**
Search for detached pipelines when doing `glab ci trace`

**Additional context**
https://docs.gitlab.com/ee/ci/merge_request_pipelines/

tommyknows:
**Description**
For `ci trace` and `ci status`, the `GetPipelines` function has been
used. However, `GetPipelines` does not with [detached pipelines](https://docs.gitlab.com/ee/ci/merge_request_pipelines/).

To fix this, we're now using a new function `GetLastPipeline`, which
first tries to get the last pipeline through the `GetCommit` API and if
that should not return any results, it falls back to the old
`GetPipelines` function.

Some parts of the codes have been refactored to use this new function
instead of listing all pipelines and then grabbing the first elementt of
that list.

`GetPipelines` is now only used within the `api` package itself, but I
kept it public for compatibility reasons.

**Related Issue**

Resolves #721

**How Has This Been Tested?**

Locally to a private instance that has a detached pipeline on a Merge Request.

**Types of changes**

- [ ] Documentation

### DIFF
--- a/api/pipeline.go
+++ b/api/pipeline.go
@@ -111,6 +111,39 @@ var GetJobs = func(client *gitlab.Client, repo string, opts *gitlab.ListJobsOpti
 	return jobs, nil
 }
 
+var GetLastPipeline = func(client *gitlab.Client, repo string, ref string) (*gitlab.PipelineInfo, error) {
+	if client == nil {
+		client = apiClient.Lab()
+	}
+
+	c, _, err := client.Commits.GetCommit(repo, ref)
+	if err != nil {
+		return nil, err
+	}
+	if c.LastPipeline != nil {
+		return c.LastPipeline, nil
+	}
+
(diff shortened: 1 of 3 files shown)


profclems:
Thanks for working on this @tommyknows...

Kindly rebase this branch

tommyknows:
done. thanks for the quick response!

profclems (approved):
LGTM!

PR #819, opened 2021-08-12
feat: fill MR desc with all commit bodies

### LINKED ISSUE #795: Add commit bodies on MRs with multiple commits

Creating an MR with multiple commits, I would like to have the commit bodies added to the MR description as well as the titles

**Propose a Solution**

Up to you whether this should be the default (same as single commit MRs) or behind an additional flag such as `--fill-body` or something. 
This can be added [here](https://github.com/profclems/glab/blob/0fb95eab621e8ea12d1a870889c6f29a72caa9aa/commands/mr/create/mr_create.go#L504) where the commit titles are currently being added.

**Additional context**

I can work on this, just wanted to get thoughts on the flag.

guyfedwards:
**Description**
adds --fill-commit-body flag to add all the commit bodies to the MR description
when creating an my with --fill

closes #795

**Related Issue**
Resolves #795

**How Has This Been Tested?**
Wrote tests for this and tested on personal gitlab account

@profclems @zemzale not sure how I feel about the formatting, ends up looking kinda weird because of the indentation

**Types of changes**

- [ ] Documentation

### DIFF
--- a/commands/mr/create/mr_create.go
+++ b/commands/mr/create/mr_create.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"net/url"
 	"os"
+	"regexp"
 	"strings"
 
 	"github.com/profclems/glab/pkg/iostreams"
@@ -40,14 +41,15 @@ type CreateOpts struct {
 	RemoveSourceBranch bool
 	AllowCollaboration bool
 
-	Autofill      bool
-	IsDraft       bool
-	IsWIP         bool
-	ShouldPush    bool
-	NoEditor      bool
-	IsInteractive bool
-	Yes           bool
-	Web           bool
+	Autofill       bool
+	FillCommitBody bool
+	IsDraft        bool
+	IsWIP          bool
+	ShouldPush     bool
+	NoEditor       bool
+	IsInteractive  bool
+	Yes            bool
+	Web            bool
 
 	IO       *iostreams.IOStreams
 	Branch   func() (string, error)
@@ -84,6 +86,7 @@ func NewCmdCreate(f *cmdutils.Factory, runE func(opts *CreateOpts) error) *cobra
 			$ glab mr create -a username -t "fix annoying bug"
 			$ glab mr create -f --draft --label RFC
 			$ glab mr create --fill --yes --web
+			$ glab mr create --fill --fill-commit-body --yes
 		`),
 		Args: cobra.ExactArgs(0),
 		PreRun: func(cmd *cobra.Command, args []string) {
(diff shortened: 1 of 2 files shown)


zemzale (dismissed):
Nice work @guyfedwards. LGTM.

The formatting maybe is not the best, but I can't think of anything better. I am for adding a feature and improving upon it later on

profclems:
> **Screenshots (if appropriate):**
> @profclems @zemzale not sure how I feel about the formatting, ends up looking kinda weird because of the indentation
> 
> 

I'm not convinced about this too. Looks really weird. To fix this in markdown, you need two spaces after the first line like so:

```
- first line <space><space>
this is the commit description
```

This will be rendered as expected:

- this is a test commit title  
this is the commit description
and a really long one.
- this is a second test commit title  
this is the commit descriptions
for the second commit

commands/mr/create/mr_create.go
```diff
@@ -501,7 +503,16 @@ func mrBodyAndTitle(opts *CreateOpts) error {
 		if opts.Description == "" {
 			var body strings.Builder
 			for i := len(commits) - 1; i >= 0; i-- {
-				fmt.Fprintf(&body, "- %s\n", commits[i].Title)
+				// adds 2 spaces for markdown line wrapping
+				fmt.Fprintf(&body, "- %s  \n", commits[i].Title)
```
profclems, line 512: Oops this rules out my comment here https://github.com/profclems/glab/pull/819#issuecomment-897382936.

Did you test with this on GitLab after adding the spaces?

commands/mr/create/mr_create.go
```diff
@@ -135,6 +136,7 @@ func NewCmdCreate(f *cmdutils.Factory, runE func(opts *CreateOpts) error) *cobra
 		},
 	}
 	mrCreateCmd.Flags().BoolVarP(&opts.Autofill, "fill", "f", false, "Do not prompt for title/description and just use commit info")
+	mrCreateCmd.Flags().BoolVarP(&opts.FillCommitBody, "fill-commit-body", "", false, "Fill description with each commit body when multiple commits")
```
profclems: More context need in the flag description.
```
	mrCreateCmd.Flags().BoolVarP(&opts.FillCommitBody, "fill-commit-body", "", false, "Fill description with each commit body when multiple commits. Can only be used with --fill")
```

And it should error if `--fill-commit-body is used without --fill`. Just like as done on line 119:

```go
if opts.FillCommitBody && !opts.Autofill {
	return &cmdutils.FlagError{Err: errors.New("--fill-commit-body should be used with --fill")}
}
```

profclems (changes requested):
Please add usage examples here: 

https://github.com/profclems/glab/blob/999d1ff79da53f25ddebb237ffdc1b3d84659c2f/commands/mr/create/mr_create.go#L83-L88

commands/mr/create/mr_create.go
guyfedwards: Ah yes, the screenshot is slightly outdated I think. The two spaces aligns the main body of the commit but the `closes  #902` is then misaligned. One option would be to replace all `\n` with double space in the body?

profclems: I think that happens because there is `\n\n` before the `closes #902`. Maybe we can replace the `\n\n` with `\n`??

profclems: Another option is to insert double spaces before each `\n`. So we replace all `\n` with `<double-space>\n`

guyfedwards: Went with a combination of both, replacing `\n\n` with `<double-space>\n`. Looks much better now, thanks for the suggestions

guyfedwards:
@profclems I have made the requested changes but github seems to be stuck thinking I haven't?

profclems (approved):

PR #845, opened 2021-09-24
feat: allow reuse of glab in other CLI contexts

### LINKED ISSUE #843: Refactor to enable packaging glab inside another golang CLI tool (overriding config location)

I would like to incorporate a `gitlab` sub command into a CLI I am developing for MediaWiki / Wikimedia developers.
This would work as part of an already released binary, and be automatically configured to point to the Wikimedia Gitlab instance

**Propose a Solution**

Refactor `glinstance` and or Factory to allow overriding the location of the configuration file as well as the domain to be use for gitlab.
This would allow users to use a version of the glab CLI packaged within the mwcli as well as a standalone glab binary pointing to the default instance.

**Additional context**

It would appear something like this:

image

Other CLI codebase: https://gitlab.wikimedia.org/releng/cli

addshore:
**Description**
Moves glinstance into pkg for use by other projects.
Uses OverridableDefault hostname in `auth login` command.

**Related Issue**
Resolves #843

**How Has This Been Tested?**

- Tested the build which still works.
- Tested the code change included in the mwcli.

Usage in the other CLI can be found in https://phabricator.wikimedia.org/T288373#7376987
Code change for the other CLI can be found at https://gitlab.wikimedia.org/releng/cli/-/merge_requests/47

**Types of changes**

- [ ] Documentation


commands/auth/login/login.go
```diff
@@ -125,14 +125,15 @@ func loginRun() error {
 
 	hostname := opts.Hostname
 	apiHostname := opts.Hostname
+	defaultHostname := glinstance.Default()
```
profclems: Use the [OverridableDefault()](https://github.com/profclems/glab/blob/c608850edb768d2a9be2d98874440a3643a32a42/internal/glinstance/host.go#L28) instead of [Default()](https://github.com/profclems/glab/blob/c608850edb768d2a9be2d98874440a3643a32a42/internal/glinstance/host.go#L18). 
Default() returns the value in the [`defaultHostname`](https://github.com/profclems/glab/blob/c608850edb768d2a9be2d98874440a3643a32a42/internal/glinstance/host.go#L10) which is constant. The value is `gitlab.com`.
OverridableDefault() will check if the value has been overriden and returns it, if not, returns the default.
```suggestion
	defaultHostname := glinstance.OverridableDefault()
```

profclems (changes requested):

commands/auth/login/login.go
profclems: Oops this has been resolved in the other commit

profclems (approved):

addshore:
Thanks!
I'll add this to the CLI fully for the next release and upstream any other tweaks that might make sense for this kind of usecase!

PR #853, opened 2021-10-03
chore: use errors.Is instead of error equality

hoelzro:
This might be needlessly pedantic, but it insulates against potential
future changes in error wrapping

**Description**
Removes all of the checks I could find that directly compare error values, rather than using `errors.Is`

**Related Issue**

N/A

**How Has This Been Tested?**

I just made sure the code compiles, plus I did a few one-off manual tests.

**Types of changes**

- [ ] Documentation
- [x] Chore (Related to CI or Packaging to platforms)


zemzale:
This is a nice small fix. Thank you @hoelzro

PR #868, opened 2021-10-07
feat(command/ci/retry): Add new command to retry CI jobs

### LINKED ISSUE #702: cli way of retrying job

I find it cumbersome having to navigate in the ci TUI in order to retry a job. Not to mention that makes totally impossible the integration with other commands or scripts.

**Propose a Solution**

`glab ci retry [<job-id-or-name>]`

**Additional context**

If/when we agree on something, I can start creating a PR

szemek:
## Description
Changes add new command that allows to retry CI jobs by job id

## Related Issue
Resolves #702

## How Has This Been Tested?
Changes were tested manually.

## Types of changes

- [ ] Documentation


commands/ci/retry/retry.go
```diff
@@ -0,0 +1,51 @@
...
+			if err != nil {
+				return err
+			}
+
+			repo, err := f.BaseRepo()
+			if err != nil {
+				return err
+			}
+
+			jobID := utils.StringToInt(args[0])
```
profclems, line 37: We should check beforehand, if the provided ID is valid. `utils.StringToInt()` returns 0 if the value isn't a valid int

```suggestion
			jobID := utils.StringToInt(args[0])
			if jobID < 1 {
				fmt.Fprintf(f.IO.StdErr, "invalid job id: %d", jobID)
				return cmdutils.SilentError
			}
```

commands/ci/retry/retry.go
```diff
@@ -0,0 +1,51 @@
...
+			job, err := api.RetryPipelineJob(apiClient, jobID, repo.FullName())
+			if err != nil {
+				fmt.Fprintf(f.IO.StdOut, "Could not retry a job with ID %d\n", jobID)
+				return err
```
profclems: You can simplify this with `return cmdutils.WrapError(err, fmt.Sprintf("Could not retry job with ID: %d", jobID))`

docs/source/ci/retry.rst
```diff
@@ -0,0 +1,33 @@
...
+  
+
+Options inherited from parent commands
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+::
+
+      --help              Show help for command
+  -R, --repo OWNER/REPO   Select another repository using the OWNER/REPO or `GROUP/NAMESPACE/REPO` format or full URL or git URL
+
```
profclems: You don't need to manually update the docs. It is auto-generate at every release

commands/ci/retry/retry.go
```diff
@@ -0,0 +1,51 @@
...
+	"github.com/profclems/glab/commands/cmdutils"
+	"github.com/profclems/glab/pkg/utils"
+
+	"github.com/MakeNowJust/heredoc"
+	"github.com/spf13/cobra"
+)
+
+func NewCmdRetry(f *cmdutils.Factory) *cobra.Command {
+	var pipelineRetryCmd = &cobra.Command{
+		Use:     "retry <id>",
```
profclems: Best to be more specify with the type of ID required. I would choose `<job-id>`.

profclems (changes requested):
Great work on this @szemek ! 

Just a few fixes. Also I was thinking we could support job names as well and not just the ID

docs/source/ci/retry.rst
szemek: Roger that. Thanks!

commands/ci/retry/retry.go
szemek: Thanks for the suggestion. I will display original args[0] instead of jobID in the output
```go
	if jobID < 1 {
		fmt.Fprintln(f.IO.StdErr, "invalid job id:", args[0])
		return cmdutils.SilentError
	}
```

profclems: > I will display original args[0] instead of jobID in the output

That's even more perfect! Thanks

profclems (approved):
LGTM! Thanks @szemek!

If you ever get time, you can work on adding support for job names and not just the ID, maybe in a separate PR.

PR #881, opened 2021-10-19
fix: allow setting variable without using --value

### LINKED ISSUE #880: Setting a variable only works when using --value flag
**Description**
The help output for `glab variable set` show that you can either provide two arguments with the first being the key and the second being the value, or you can provide the key name as an argument and use the `--value/-v` flag to specify the value. However, if you don't provide the `--value` flag, the variable is created but the value is empty.

- `glab variable set WITH_ARG "some value"` should create a variable named `WITH_ARG` that has the value `some value`.
- `glab variable set WITH_ARG "some value"` creates a variable named `WITH_ARG` and the value is empty.

**Steps to Reproduce**
1. Run `glab variable set WITH_ARG "some value"`
2. View the variable in the gitlab UI and you'll see it is set to an empty value

**Your Environment**
* Version used (Run `glab --version`):

```
$ glab --version
glab version 1.21.1 (2021-09-28)
```

```
OS: Ubuntu 20.04.3 LTS x86_64 
Kernel: 5.13.0-7614-generic 
Shell: /bin/zsh 5.8 
```

bradym:
## Description
This PR addresses #880 by simplifying and consolidating the logic for setting `opts.Value`.

Currently if there are two args to the `glab variable set` command, `opts.Value` [is set to the second arg](https://github.com/profclems/glab/blob/0d47c686086a27da3ac1543db4a63587ea5f1c34/commands/variable/set/set.go#L74). Then later the `getValue` function is called, and further changes are made to the value.

This update:

- Removes the [redundant if statement](https://github.com/profclems/glab/blob/0d47c686086a27da3ac1543db4a63587ea5f1c34/commands/variable/set/set.go#L70)
- Removes the setting of opts.value [outside of the `getValue` function](https://github.com/profclems/glab/blob/0d47c686086a27da3ac1543db4a63587ea5f1c34/commands/variable/set/set.go#L74)
- Removes the [use of opts.ValueSet](https://github.com/profclems/glab/blob/0d47c686086a27da3ac1543db4a63587ea5f1c34/commands/variable/set/set.go#L82)
- Updates the logic in the `getValue` function to enable setting the variable with or without passing `--value` flag

## Related Issue
Resolves #880 

## How Has This Been Tested?

I ran all of the commands from the help command for `glab variable set` before and after my change:
```
$ glen | sort -r
export ORIGINAL_WITH_ARG=""
export ORIGINAL_SERVER_TOKEN="SERVER_TOKEN_VALUE"
export ORIGINAL_FROM_FLAG="some value"
export ORIGINAL_FROM_FILE="FROM_FILE_VALUE"
export ORIGINAL_FROM_ENV_WITH_FLAG="bradym"
export ORIGINAL_FROM_ENV_WITH_ARG=""
export FIXED_WITH_ARG="some value"
export FIXED_SERVER_TOKEN="SERVER_TOKEN_VALUE"
export FIXED_FROM_FLAG="some value"
export FIXED_FROM_FILE="FROM_FILE_VALUE"
export FIXED_FROM_ENV_WITH_FLAG="bradym"
export FIXED_FROM_ENV_WITH_ARG="bradym"
```
I also ran tests in a docker container using golang 1.16:

```
docker run --rm -ti --init -v "$PWD:/app" -w /app golang:1.15-buster bash
root@6c4876a30b19:/app# bin/gotestsum --packages ./commands/variable/set
✓  commands/variable/set (cached)

DONE 18 tests in 0.255s
```
I tried running `make test` but:

```
DONE 847 tests, 10 failures in 610.099s
```
Reviewing the errors it looks like it's due to permissions issues. 

## Types of changes

- [ ] Documentation

### DIFF
--- a/commands/variable/set/set.go
+++ b/commands/variable/set/set.go
@@ -29,8 +29,6 @@ type SetOpts struct {
 	Protected bool
 	Masked    bool
 	Group     string
-
-	ValueSet bool
 }
 
 func NewCmdSet(f *cmdutils.Factory, runE func(opts *SetOpts) error) *cobra.Command {
@@ -67,11 +65,8 @@ func NewCmdSet(f *cmdutils.Factory, runE func(opts *SetOpts) error) *cobra.Comma
 			}
 
 			if opts.Value != "" && len(args) == 2 {
-				if opts.Value != "" {
-					err = cmdutils.FlagError{Err: errors.New("specify value either by second positional argument or --value flag")}
-					return
-				}
-				opts.Value = args[1]
+				err = cmdutils.FlagError{Err: errors.New("specify value either by second positional argument or --value flag")}
+				return
 			}
 
 			if cmd.Flags().Changed("scope") && opts.Group != "" {
@@ -79,8 +74,7 @@ func NewCmdSet(f *cmdutils.Factory, runE func(opts *SetOpts) error) *cobra.Comma
 				return
 			}
 
-			opts.ValueSet = cmd.Flags().Changed("value") || len(args) == 2
-			opts.Value, err = getValue(opts)
+			opts.Value, err = getValue(opts, args)
 			if err != nil {
 				return
 			}
@@ -156,9 +150,11 @@ func setRun(opts *SetOpts) error {
 	return nil
 }
 
-func getValue(opts *SetOpts) (string, error) {
-	if opts.Value != "" || opts.ValueSet {
+func getValue(opts *SetOpts, args []string) (string, error) {
+	if opts.Value != "" {
 		return opts.Value, nil
+	} else if len(args) == 2 {
+		return args[1], nil
 	}
 
 	if opts.IO.IsInTTY {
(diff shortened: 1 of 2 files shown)


commands/variable/set/set.go
```diff
@@ -156,9 +150,13 @@ func setRun(opts *SetOpts) error {
 	return nil
 }
 
-func getValue(opts *SetOpts) (string, error) {
-	if opts.Value != "" || opts.ValueSet {
+func getValue(opts *SetOpts, args []string) (string, error) {
+	if opts.Value != "" {
 		return opts.Value, nil
+	} else {
+		if len(args) == 2 {
```
profclems: The nested if can be replaced with else if

profclems (changes requested):
Thanks for working on this @bradym

commands/variable/set/set.go
bradym: Oops, good catch. Will push a fix for this.

profclems (approved):

PR #883, opened 2021-10-19
feat(mr): Only output MR URL to stdout if output is not a TTY

carlpett:
This enables "glab mr create -f -y | some command" to only pass the url
through the pipe.

### DIFF
--- a/commands/mr/mrutils/mrutils.go
+++ b/commands/mr/mrutils/mrutils.go
@@ -84,10 +84,13 @@ func MRCheckErrors(mr *gitlab.MergeRequest, err MRCheckErrOptions) error {
 	return nil
 }
 
-func DisplayMR(c *iostreams.ColorPalette, mr *gitlab.MergeRequest) string {
+func DisplayMR(c *iostreams.ColorPalette, mr *gitlab.MergeRequest, isTTY bool) string {
 	mrID := MRState(c, mr)
-	return fmt.Sprintf("%s %s (%s)\n %s\n",
-		mrID, mr.Title, mr.SourceBranch, mr.WebURL)
+	if isTTY {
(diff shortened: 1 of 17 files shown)


commands/mr/mrutils/mrutils.go
```diff
@@ -84,10 +84,13 @@ func MRCheckErrors(mr *gitlab.MergeRequest, err MRCheckErrOptions) error {
 	return nil
 }
 
-func DisplayMR(c *iostreams.ColorPalette, mr *gitlab.MergeRequest) string {
+func DisplayMR(c *iostreams.ColorPalette, mr *gitlab.MergeRequest, isTTY bool) string {
```
carlpett, line 87: I'm not super happy about this signature, to be honest, but since you already passed around the `IsaTTY` I made the guess that it'd also be at odds with the design to just import the detection library here?

profclems (commented):
To make things even, `issue` command should also have this behaviour.

carlpett:
Good point, thank @profclems. Added this now.

commands/mr/mrutils/mrutils.go
zemzale: I agree that the signature is not the best. Even the deepsourse analyzes say exactly that. 

But I think this is fine, for what we are doing. If we would expand over these values I would say that we should look for a better solution, but for now this tech debt is acceptable IMO.

zemzale: @profclems I don't know if it's possible to ignore the deepsource analyzer, but what are your thoughts on passing around the `isTTY` value?

profclems: We can live with that for now

profclems (approved):

PR #886, opened 2021-10-21
fix: Fail with a specific error when unauthenticated

### LINKED ISSUE #792: no gitlab projects found from remotes
**Description**
Getting the error "no gitlab projects found from remotes" when running glab issue list
`glab auth status` returns all green check marks
List the issues

**Steps to Reproduce**
$ glab auth login --stdin < token.txt
$ glab issue list

**Your Environment**
glab version 1.18
gitlab version 14.01

### LINKED ISSUE #877: Can't get glab to work in repo
**Description**
I started to experiment around with glab, but somehow i can't get past the repo clone:

``` shell
~ # glab auth login -h ${GITLAB_HOST} -t ${GITLAB_TOKEN}
~ # glab repo clone "https://${GITLAB_HOST}/EDV/dokumentation"
Cloning into 'dokumentation'...
remote: Enumerating objects: 902, done.
remote: Counting objects: 100% (15/15), done.
remote: Compressing objects: 100% (12/12), done.
remote: Total 902 (delta 5), reused 10 (delta 3), pack-reused 887
Receiving objects: 100% (902/902), 1.97 MiB | 10.23 MiB/s, done.
Resolving deltas: 100% (518/518), done.
~ # cd dokumentation/
~/dokumentation # glab issue list
no GitLab Projects found from remotes
```

am i doing something wronge? I searched the whole documentation but could not find a hint about what to do here!

carlpett:
If resolutions fails, return the error instead of "no GitLab Projects found from remotes"

## Description
Currently, a misleading/confusing error is returned if authentication fails. This PR changes it to return the actual error instead.

## Related Issue

_Possibly_ resolves #792
_Possibly_ resolves #877

## How Has This Been Tested?
Tested with bad token, added some test cases

## Screenshots (if appropriate)
-

## Types of changes

- [ ] Documentation

### DIFF
--- a/internal/glrepo/resolver.go
+++ b/internal/glrepo/resolver.go
@@ -2,6 +2,7 @@ package glrepo
 
 import (
 	"errors"
+	"fmt"
 	"sort"
 	"strings"
 
@@ -9,6 +10,7 @@ import (
 	"github.com/profclems/glab/pkg/git"
 	"github.com/profclems/glab/pkg/prompt"
 
+	"github.com/hashicorp/go-multierror"
 	"github.com/xanzy/go-gitlab"
 )
 
@@ -39,15 +41,21 @@ func ResolveRemotesToRepos(remotes Remotes, client *gitlab.Client, base string) 
 
 func resolveNetwork(result *ResolvedRemotes) error {
 	// Loop over at most 5 (maxRemotesForLookup)
+	var errs error
+	anySuccess := false
 	for i := 0; i < len(result.remotes) && i < maxRemotesForLookup; i++ {
 		networkResult, err := api.GetProject(result.apiClient, result.remotes[i].FullName())
 		if err == nil {
 			result.network = append(result.network, *networkResult)
+			anySuccess = true
 		} else {
-			return err
+			errs = multierror.Append(errs, fmt.Errorf("%s: %w", result.remotes[i].FullName(), err))
 		}
 	}
-	return nil
+	if anySuccess {
+		return nil
+	}
+	return errs
 }
 
 type ResolvedRemotes struct {
--- a/internal/glrepo/resolver.go
+++ b/internal/glrepo/resolver.go
@@ -37,14 +37,17 @@ func ResolveRemotesToRepos(remotes Remotes, client *gitlab.Client, base string) 
 	return result, nil
 }
 
-func resolveNetwork(result *ResolvedRemotes) {
+func resolveNetwork(result *ResolvedRemotes) error {
 	// Loop over at most 5 (maxRemotesForLookup)
 	for i := 0; i < len(result.remotes) && i < maxRemotesForLookup; i++ {
 		networkResult, err := api.GetProject(result.apiClient, result.remotes[i].FullName())
 		if err == nil {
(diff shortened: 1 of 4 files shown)


carlpett:
@profclems Anything I need to do to get this reviewed, or just a question of finding time on your end? The DeepSource thing does not look related to my changes, unsure what to do about it.

zemzale (commented):
This change would likely break network resolving for someone. 

The current implementation loops over the remotes(max number of 5) and adds them to the networks list, if there is no error.

The new implementation would fail with the first error(which could be any not 200 OK response from GitLab), meaning it would not try to connect to all possible remotes.

The idea to add more context to the error is good, but the implementation needs something more. 
I imagine maybe saving the errors into a slice, and if no remotes got found, then create one big error that gets passed down the stack with all the context?

@profclems some extra info would be great here, since I have not deep dived into the interlans of the network resolver.

carlpett:
So, a `network` is a remote? I think I was a bit confused on what it was meant to represent.
Would the correct behaviour be to succeed if _any_ remote returns successfully?

zemzale:
I am gonna be honest, but those namings confuse me, but yeah the correct behavior would be to succeed if any remote returns successfully. 

This whole code path basically is our way of trying to find out to which remotes for this project might be available.

profclems:
Hi @carlpett, @zemzale is right. This change will break network resolution for repositories with multiple authenticated remotes. 

The idea behind the network resolution is to find the remote that is authenticated among the available remotes. This will fail if any of the remotes is unreachable. 

Maybe we can group the errors and print them out when the resolution fails just like @zemzale said.

internal/glrepo/resolver.go
```diff
@@ -37,14 +37,17 @@ func ResolveRemotesToRepos(remotes Remotes, client *gitlab.Client, base string)
...
 }
 
-func resolveNetwork(result *ResolvedRemotes) {
+func resolveNetwork(result *ResolvedRemotes) error {
 	// Loop over at most 5 (maxRemotesForLookup)
 	for i := 0; i < len(result.remotes) && i < maxRemotesForLookup; i++ {
 		networkResult, err := api.GetProject(result.apiClient, result.remotes[i].FullName())
 		if err == nil {
 			result.network = append(result.network, *networkResult)
+		} else {
```
profclems, line 51: Seems like we are taking the fail fast approach and this will fail network resolution if the first remote is not reachable. 

See my comment here: https://github.com/profclems/glab/pull/886#issuecomment-959662401

profclems (changes requested):

carlpett:
I've pushed an updated version now, which only fails if all remotes fail. I opted to use hashicorp/go-multierror to do error aggregation since it was already an indirect dependency. Let me know if you have other preferences!

zemzale (approved):
This LGTM. If @profclems doesn't have any more comments I think this should be merged.

profclems (changes requested):
I think you need to run `go mod tidy` on this change

carlpett:
Ouch, missed the last comment, sorry!
Pushed a tidy

profclems (approved):

PR #887, opened 2021-10-23
added push mirror and command

NikhilMJagtap:
feat adding command for push and pull mirrors of repository

commits:
- added command for mirror
- added push mirror
- added pull mirror


commands/project/mirror/project_mirror.go
```diff
@@ -0,0 +1,128 @@
...
+
+func NewCmdMirror(f *cmdutils.Factory) *cobra.Command {
+	opts := MirrorOptions{
+		IO: f.IO,
+	}
+
+	var projectMirrorCmd = &cobra.Command{
+		Use:   "mirror [flags]",
+		Short: "Mirror a project/repository",
+		Long:  `Mirrors a project/repository to the specified location using speicified method.`,
```
profclems: Typo in ~speicified~ specified method.

Or maybe we can say: `Mirrors a project/repository to the specified location using pull or push method.` ?

api/mirror.go
```diff
@@ -0,0 +1,52 @@
...
+		URL:                   &url,
+		Enabled:               &enabled,
+		OnlyProtectedBranches: &onlyProtectedBranches,
+		KeepDivergentRefs:     &keepDivergentRefs,
+	}
+	pm, _, err := client.ProjectMirrors.AddProjectMirror(projectID, opt)
+	return pm, err
+}
+
+var CreatePullMirror = func(
```
profclems, line 38: You can enable pull mirroring using the Edit Project API https://docs.gitlab.com/ee/api/projects.html#edit-project which should be available in `go-gitlab`.

commands/project/mirror/project_mirror.go
```diff
@@ -0,0 +1,152 @@
...
+		Args:  cobra.MaximumNArgs(1),
```
profclems, line 39: Why not `cobra.ExactArgs(1)` ? Also, if we are using cobra.MaximumNArgs(1), then we should allow no argument at all and use the current local git repo from which the command is executed.

profclems (changes requested):
This is great @NikhilMJagtap!

commands/project/mirror/project_mirror.go
NikhilMJagtap: `ExactArgs` is better option, I agree.

commands/project/mirror/project_mirror.go
NikhilMJagtap: Noted. Thanks.

api/mirror.go
```diff
@@ -0,0 +1,45 @@
...
+	"github.com/xanzy/go-gitlab"
+)
+
+var CreatePushMirror = func(
+	client *gitlab.Client,
+	projectID interface{},
+	url string,
+	enabled bool,
+	onlyProtectedBranches bool,
+	keepDivergentRefs bool,
```
profclems: we can have these as separate type struct options for CreatePushMirror
```go
type CreatePushMirrorOptions struct {
    url string
    enabled bool
    onlyProtectedBranches bool
    keepDivergentRefs bool
}
```
And the function signature becomes: 
```go
var CreatePushMirror = func(client *gitlab.Client, projectID interface{}, opts *CreatePushMirrorOptions) (*gitlab.ProjectMirror, error)
```

api/mirror.go
```diff
@@ -0,0 +1,45 @@
...
+	pm, _, err := client.ProjectMirrors.AddProjectMirror(projectID, opt)
+	return pm, err
+}
+
+var CreatePullMirror = func(
+	client *gitlab.Client,
+	projectID interface{},
+	url string,
+	enabled bool,
+	onlyProtectedBranches bool,
```
profclems: Same as above ^^

commands/project/mirror/project_mirror.go
```diff
@@ -0,0 +1,145 @@
...
+			}
+			opts.ProjectID = project.ID
+			return runProjectMirror(f, &opts)
+		},
+	}
+	projectMirrorCmd.Flags().StringVar(&opts.URL, "url", "", "The target URL to which the repository is mirrored.")
+	projectMirrorCmd.Flags().StringVar(&opts.Direction, "direction", "pull", "Mirror direction")
+	projectMirrorCmd.Flags().BoolVar(&opts.Enabled, "enabled", true, "Determines if the mirror is enabled.")
+	projectMirrorCmd.Flags().BoolVar(&opts.ProtectedBranchesOnly, "protected-branches-only", false, "Determines if only protected branches are mirrored.")
+	projectMirrorCmd.Flags().BoolVar(&opts.AllowDivergence, "allow-divergence", false, "Determines if divergent refs are skipped.")
```
profclems, line 107: We should check that the user does not use `--direction pull` and `--allow-divergence` together, since the `--allow-divergence` is useless for pull mirroring.

commands/project/mirror/project_mirror.go
```diff
@@ -0,0 +1,145 @@
...
+			}
+
+			opts.URL = strings.TrimSpace(opts.URL)
+
+			if opts.URL == "" {
+				return cmdutils.WrapError(
+					errors.New("argument URL is required"),
+					"argument URL must be provided",
+				)
+			}
```
profclems: We may have to reword this error message. `url` is a flag, not an argument.

Or the easiest thing to do is mark the flag as required with so you don't have to manually validate.

Here is an example: https://github.com/profclems/glab/blob/045610825ba1362f80524f7be3ccb47d059d4a09/commands/ssh-key/add/add.go#L69-L74

commands/project/mirror/project_mirror.go
```diff
@@ -0,0 +1,145 @@
...
+			project, err := api.GetProject(apiClient, opts.ProjectName)
+			if err != nil {
+				return cmdutils.WrapError(err, "project/repository not found")
+			}
+			opts.ProjectID = project.ID
+			return runProjectMirror(f, &opts)
+		},
+	}
+	projectMirrorCmd.Flags().StringVar(&opts.URL, "url", "", "The target URL to which the repository is mirrored.")
+	projectMirrorCmd.Flags().StringVar(&opts.Direction, "direction", "pull", "Mirror direction")
```
profclems, line 104: Should we avoid the default and mark this flag too as required?

profclems (changes requested):

profclems (approved):
Looks great! Thanks for working on this @NikhilMJagtap.

Note that I will be adding a few commits to allow for full gitlab project url instead of just `project/repository`

NikhilMJagtap:
@profclems 
Any update on this? Getting marked stale so pinging you once. Please ignore if expected or normal. Thanks.

PR #895, opened 2021-11-03
fix: add linebreak to check-output msg

Spreadcat:
## Description

This fix adds a linebreak after the check-update output message.

## Related Issue

none

## Types of changes

### DIFF
--- a/commands/update/check_update.go
+++ b/commands/update/check_update.go
@@ -43,7 +43,7 @@ func CheckUpdate(s *iostreams.IOStreams, version string, silentErr bool) error {
 			return nil
 		}
 		fmt.Fprintf(s.StdOut, "%v %v", c.GreenCheck(),
-			c.Green("You are already using the latest version of glab"))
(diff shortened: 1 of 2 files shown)


commands/update/check_update.go
```diff
@@ -43,7 +43,7 @@ func CheckUpdate(s *iostreams.IOStreams, version string, silentErr bool) error {
 			return nil
 		}
 		fmt.Fprintf(s.StdOut, "%v %v", c.GreenCheck(),
-			c.Green("You are already using the latest version of glab"))
+			c.Green("You are already using the latest version of glab\n"))
```
profclems, line 46: Or maybe we could add the line break to the second param
```suggestion
		fmt.Fprintf(s.StdOut, "%v %v\n", c.GreenCheck(),
			c.Green("You are already using the latest version of glab"))
```

Spreadcat: I can totally live with that.

PR #910, opened 2021-11-22
Add artifact download command

### LINKED ISSUE #888: Download artifacts

Now, ci and pipeline are supported to check status and log. Sometimes, a job may produce some artifacts. Can it be downloaded by this tool?

Simply, glab can generate a url from [get job artifacts](https://docs.gitlab.com/ee/api/job_artifacts.html) and pass it to curl.

jpeedroza:
## Description
This changes create a new command to download all files inside the Artifact from the last pipeline.

## Related Issue
Resolves #888 

## How Has This Been Tested?
I tested this command with my own repository, generating the artifact from the pipeline with various directories and files.

## Types of changes
- [X] New feature (non-breaking change which adds functionality)
- [ ] Documentation
- [X] Chore (Related to CI or Packaging to platforms)

### DIFF
--- a/commands/ci/artifact/artifact.go
+++ b/commands/ci/artifact/artifact.go
@@ -38,24 +38,38 @@ func NewCmdRun(f *cmdutils.Factory) *cobra.Command {
 				return err
 			}
 
-			artifact, _ := api.DownloadArtifactJob(apiClient, repo.FullName(), args[0], &gitlab.DownloadArtifactsFileOptions{Job: &args[1]})
+			artifact, err := api.DownloadArtifactJob(apiClient, repo.FullName(), args[0], &gitlab.DownloadArtifactsFileOptions{Job: &args[1]})
+			if err != nil {
+				return err
+			}
 
-			zr, _ := zip.NewReader(artifact, artifact.Size())
-			os.Mkdir(p, 0777)
+			zr, err := zip.NewReader(artifact, artifact.Size())
+			if err != nil {
+				return err
+			}
+
+			if err := os.Mkdir(p, 0755); err != nil {
+				return err
+			}
+
 			for _, v := range zr.File {
 				if v.FileInfo().IsDir() {
-					os.Mkdir(p+v.Name, 0777)
+					if err := os.Mkdir(p+v.Name, v.Mode()); err != nil {
+						return err
+					}
 				} else {
-					srcFile, _ := zr.Open(v.Name)
+					srcFile, err := zr.Open(v.Name)
+					if err != nil {
+						return err
+					}
 					defer srcFile.Close()
 					dstFile, err := os.OpenFile(p+v.Name, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, v.Mode())
 					if err != nil {
-						panic(err)
+						return err
 					}
 					io.Copy(dstFile, srcFile)
 				}
 			}
-
 			return nil
 		},
 	}
--- a/commands/ci/artifact/artifact.go
+++ b/commands/ci/artifact/artifact.go
@@ -1,0 +1,65 @@
+package ci
+
+import (
+	"archive/zip"
+	"io"
+	"os"
+
+	"github.com/MakeNowJust/heredoc"
+	"github.com/profclems/glab/api"
+	"github.com/profclems/glab/commands/cmdutils"
+	"github.com/spf13/cobra"
+	"github.com/xanzy/go-gitlab"
+)
+
+func NewCmdRun(f *cmdutils.Factory) *cobra.Command {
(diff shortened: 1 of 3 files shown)


commands/ci/artifact/artifact.go
```diff
@@ -0,0 +1,65 @@
...
+			}
+			p, err := cmd.Flags().GetString("path")
+			if err != nil {
+				return err
+			}
+
+			artifact, _ := api.DownloadArtifactJob(apiClient, repo.FullName(), args[0], &gitlab.DownloadArtifactsFileOptions{Job: &args[1]})
+
+			zr, _ := zip.NewReader(artifact, artifact.Size())
+			os.Mkdir(p, 0777)
```
zemzale: The error from `os.Mkdir` shouldn't be ignored. You need to check it and return it if something goes wrong.

Also, the rights of 777 are not correct here. I think something like 644 is more suitable.

commands/ci/artifact/artifact.go
```diff
@@ -0,0 +1,65 @@
...
+			for _, v := range zr.File {
+				if v.FileInfo().IsDir() {
+					os.Mkdir(p+v.Name, 0777)
```
zemzale: The error from `os.Mkdir` shouldn't be ignored. You need to check it and return it if something goes wrong.

Also, the rights of 777 are not correct here. I think something like 644 is more suitable.

commands/ci/artifact/artifact.go
```diff
@@ -0,0 +1,65 @@
...
+			apiClient, err := f.HttpClient()
+			if err != nil {
+				return err
+			}
+			p, err := cmd.Flags().GetString("path")
+			if err != nil {
+				return err
+			}
+
+			artifact, _ := api.DownloadArtifactJob(apiClient, repo.FullName(), args[0], &gitlab.DownloadArtifactsFileOptions{Job: &args[1]})
```
zemzale: The error shouldn't be ignored over here. It should be checked and returned

commands/ci/artifact/artifact.go
```diff
@@ -0,0 +1,65 @@
...
+
+			zr, _ := zip.NewReader(artifact, artifact.Size())
```
zemzale: The error shouldn't be ignored over here. It should be checked and returned

commands/ci/artifact/artifact.go
```diff
@@ -0,0 +1,65 @@
...
+			os.Mkdir(p, 0777)
+			for _, v := range zr.File {
+				if v.FileInfo().IsDir() {
+					os.Mkdir(p+v.Name, 0777)
+				} else {
+					srcFile, _ := zr.Open(v.Name)
```
zemzale: The error shouldn't be ignored over here. It should be checked and returned

commands/ci/artifact/artifact.go
```diff
@@ -0,0 +1,65 @@
...
+					defer srcFile.Close()
+					dstFile, err := os.OpenFile(p+v.Name, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, v.Mode())
+					if err != nil {
+						panic(err)
```
zemzale: You can just return here, instead of throwing a panic.

zemzale (changes requested):
Looks good, I just have some comments on error management.

commands/ci/artifact/artifact.go
jpeedroza: i tried using the 644, but when it tries to write a file it says it doesn't have permission. The minimum directory permission I was able to get it to work was 755

zemzale: Yeah, that would make sense. I sometimes forget that you need executions rights for folders.

commands/ci/artifact/artifact.go
```diff
@@ -0,0 +1,79 @@
...
+					srcFile, err := zr.Open(v.Name)
+					if err != nil {
+						return err
+					}
+					defer srcFile.Close()
+					dstFile, err := os.OpenFile(p+v.Name, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, v.Mode())
+					if err != nil {
+						return err
+					}
+					io.Copy(dstFile, srcFile)
```
profclems: We should check and return the error return value of `io.Copy`

profclems (changes requested):
Thanks for working on this @jpeedroza! I'm so excited about this.

Just a few fixes needed before we merge

commands/ci/artifact/artifact.go
zemzale: Resolving since the change has been made

profclems (approved):

PR #921, opened 2021-11-28
feature: Read config directory from GLAB_CONFIG_DIR

kampka:
## Description

When working with many different gitlab instances for different
projects, the global config quickly gets cluttered up with
specific settings that do not clean up conveniently. On the other hand,
local configs are to local, restricting the config to one repository
instead of one project or instance that potentially contains many
relevant repositories.
This change introduces a middle ground solution via the GLAB_CONFIG_DIR
environment variable. Setting this variable will allow global
configuration to travel with the project instead of populating the XDG
global config.

- [ ] Documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -293,7 +293,7 @@ Get a GitLab access token at <https://gitlab.com/-/profile/personal_access_token
 
 ## Configuration
 
-`glab` follows the XDG Base Directory [Spec](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html): global configuration file is saved at `~/.config/glab-cli`. Local configuration file is saved at `.git/glab-cli` in the current working git directory.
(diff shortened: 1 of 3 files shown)


zemzale (changes requested):
I actually like this change. There should be some changes to documentation so users know about this.

Also would need approval from @profclems but this LGTM!

kampka:
> There should be some changes to documentation so users know about this.

Agreed. I added a comment to the README as I could not find any mentioning of the config files in the documentation at all.

profclems (changes requested):
This is great!

Please add it to the env vars here https://github.com/profclems/glab/blob/9b44934c07d1a17d19fd26d5886a752d7cf06a30/commands/root.go#L41-L60

kampka:
Done.

profclems (approved):

PR #927, opened 2021-12-14
fix: Don't show non-templates when asking chosing

### LINKED ISSUE #923: Ignore dotfiles when listing templates
Files starting with a `.` in their name should be ignored when listing available templates for issues and MRs:

```console
$ glab issue create
? Choose a template  [Use arrows to move, type to filter]
  .gitkeep
  template1
  template2
> template3
  Open a blank Issue
```

Possible code location: https://github.com/profclems/glab/blob/e866496799db83a5a5ef939ed78509f3775167c1/commands/cmdutils/cmdutils.go#L85-L88

zemzale:
## Description

In cases where users have non template files in the .gitlab/[template]/
folder they might will be picked up and displayed.

This fixes that by filtering out all the files that are hidden and don't
end in .md

## Related Issue

Resolves #923

## How Has This Been Tested?

Unit tests

## Types of changes

- [ ] Documentation

### DIFF
--- a/commands/cmdutils/cmdutils_test.go
+++ b/commands/cmdutils/cmdutils_test.go
@@ -13,6 +13,7 @@ import (
 	"github.com/profclems/glab/pkg/git"
 	"github.com/profclems/glab/pkg/prompt"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	"github.com/xanzy/go-gitlab"
 )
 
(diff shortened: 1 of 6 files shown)


zemzale:
I don't know if this is the correct way how to setup testing for this, but it works 😄

profclems (changes requested):
This looks good but I'm only concerned about the new test files in the main issue templates folder.

Maybe we could move them to `test/testdata/.gitlab` ?

zemzale:
I moved the test files to the testdata directory. 

I had to override the git function for getting top-level directory, but this should be fine as long as we don't allow overriding this outside the tests.

profclems:
> I moved the test files to the testdata directory.
> 
> I had to override the git function for getting top-level directory, but this should be fine as long as we don't allow overriding this outside the tests.

Sounds good.

commands/cmdutils/cmdutils_test.go
```diff
@@ -1350,3 +1351,33 @@ func Test_ConfirmSubmission(t *testing.T) {
...
+	tests := []struct {
+		name          string
+		give          string
+		wantTemplates []string
+		wantErr       bool
+	}{
+		{
+			name:          "Get all the issues templates",
+			give:          "issue_templates",
+			wantTemplates: []string{"Feature Request", "Bug"},
```
profclems: I think it should be `[]string{"Bug", "Feature Request"}` instead.

profclems (changes requested):
One last fix for the tests

commands/cmdutils/cmdutils_test.go
zemzale: So they show up in alphabetic order? I am just gonna sort the list before returning it.

profclems (approved):

PR #929, opened 2021-12-16
Add transfer command

### LINKED ISSUE #925: Add command for transferring repositories

Transferring repositories to a new namespace is cumbersome using the Gitlab UI, especially when it's not just a single one. As a result, it'd be nice if `glab` had the ability to automate this.

**Propose a Solution**

I've got a working prototype implemented [in this branch](https://github.com/profclems/glab/compare/trunk...x3ro:add-transfer-command). The contribution guidelines state I should add an issue before opening a PR, so here I am :)

**Additional context**

Since transferring a repositories is an operation that cannot be undone, I've added a long warning message as well as a confirmation prompt (see below, typo is already fixed). Feedback is welcome:

Screenshot 2021-12-12 at 18 48 08

x3ro:
## Description

I've added a command that allows transferring repositories to other namespaces.

## Related Issue

Resolves #925

## How Has This Been Tested?

I've only manually tested my changes so far, with repositories on gitlab.com. 

## Screenshots (if appropriate)

Running `glab repo transfer fresskoma/iolite-mock --target-namespace ljenss/Museum` would yield

Screenshot 2021-12-12 at 18 48 08

## Types of changes

- [ ] Documentation

## TODOs

- [x] Add proper descriptions of the command
- [ ] ~Update documentation~
- [ ] ...

### DIFF
--- a/commands/project/transfer/project_transfer.go
+++ b/commands/project/transfer/project_transfer.go
@@ -1,0 +1,91 @@
+package transfer
+
+import (
+	"fmt"
+
+	"github.com/MakeNowJust/heredoc"
+	"github.com/profclems/glab/commands/cmdutils"
+	"github.com/spf13/cobra"
+	"github.com/xanzy/go-gitlab"
+)
+
+func NewCmdTransfer(f *cmdutils.Factory) *cobra.Command {
+	var repoTransferCmd = &cobra.Command{
+		Use:   "transfer <command> [flags]",
+		Short: `Transfer a repository to a new namespace.`,
+		Example: heredoc.Doc(`
+			$ glab repo transfer profclems/glab --target-namespace notprofclems
+		`),
+		Args: cobra.MaximumNArgs(2),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			var err error
+
+			if len(args) != 0 {
+				err = f.RepoOverride(args[0])
+				if err != nil {
+					return err
+				}
+			}
+
+			apiClient, err := f.HttpClient()
+			if err != nil {
+				return err
+			}
+
+			repo, err := f.BaseRepo()
+			if err != nil {
+				return err
+			}
+
+			dontPromptForConfirmation, err := cmd.Flags().GetBool("danger-cannot-be-undone")
+			if err != nil {
+				return err
+			}
+
+			targetNamespace, err := cmd.Flags().GetString("target-namespace")
+			if err != nil {
+				return err
+			}
+
+			c := f.IO.Color()
+			fmt.Printf(heredoc.Doc(`
+				🔴 Danger 🔴
+
+				The operation you are about to perform is potentially irreversible.
+				You will lose control of the repository you are transferring in case you do not
+				have access to the target namespace. In addition, you won't be able to transfer
+				the repository back to the original namespace unless you have administrative access
+				to the target namespace.
+
+				Source repository: %s
+				Target namespace: %s
+
+			`), c.Yellow(repo.FullName()), c.Yellow(targetNamespace))
+
+			if !dontPromptForConfirmation {
+				err = cmdutils.ConfirmTransfer()
+				if err != nil {
+					return fmt.Errorf("unable to confirm: %w", err)
+				}
+			}
+
+			opt := &gitlab.TransferProjectOptions{}
+			opt.Namespace = targetNamespace
+
+			project, _, err := apiClient.Projects.TransferProject(repo.FullName(), opt)
+			if err != nil {
+				return err
+			}
+
+			fmt.Fprintf(f.IO.StdOut, "%s Successfully transferred repository %s to %s\n",
+				c.GreenCheck(), c.Yellow(repo.FullName()), c.Yellow(project.PathWithNamespace))
+
+			return nil
+		},
+	}
+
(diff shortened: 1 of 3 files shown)


profclems:
@x3ro thanks for working on this! This is great. Note that you don't need to manually update the docs. 
They are auto-generated at each release.

commands/project/transfer/project_transfer.go
```diff
@@ -0,0 +1,95 @@
...
+
+			greenCheck := c.Green("✓")
+			fmt.Fprintf(f.IO.StdOut, "%s Successfully transferred repository %s to %s\n",
+				greenCheck, c.Yellow(repo.FullName()), c.Yellow(project.PathWithNamespace))
+
+			return nil
+		},
+	}
+
+	repoTransferCmd.Flags().BoolP("danger-cannot-be-undone", "", false, "TODO")
```
profclems: We should use either `--force | -f` or `--yes | -y` instead of `--danger-cannot-be-undone`.

The description could be "Skip confirmation prompt and force transfer operation"

```suggestion
	repoTransferCmd.Flags().BoolP("force", "f", false, "Skip confirmation prompt and force transfer operation")
```

commands/project/transfer/project_transfer.go
```diff
@@ -0,0 +1,95 @@
...
+
+			opt := &gitlab.TransferProjectOptions{}
+			opt.Namespace = targetNamespace
+
+			project, _, err := apiClient.Projects.TransferProject(repo.FullName(), opt)
+			if err != nil {
+				return err
+			}
+
+			greenCheck := c.Green("✓")
```
profclems: Use `c.GreenCheck()` instead

x3ro:
Thanks for the review @profclems :) Regarding your change to the `force` flag, I personally would prefer at least calling it something like `danger-force`. I realize this _sounds_ silly, but in this context that might be a good thing. What I'd like to prevent with this is that people lose access to their repositories, because if you pass `--force` you wouldn't be presented with the warning dialog. What do you think?

x3ro:
ping @profclems

profclems:
> Thanks for the review @profclems :) Regarding your change to the `force` flag, I personally would prefer at least calling it something like `danger-force`. I realize this _sounds_ silly, but in this context that might be a good thing. What I'd like to prevent with this is that people lose access to their repositories, because if you pass `--force` you wouldn't be presented with the warning dialog. What do you think?

I get your point but we need to stay consistent with the other commands like `repo delete`. I would prefer we use a `-y | --yes` flag like we did for `repo delete`

x3ro:
Okay, I've changed the flag to `-y | --yes`. Are there any other changes you would like to see?

commands/project/transfer/project_transfer.go
```diff
@@ -0,0 +1,91 @@
...
+			if err != nil {
+				return err
+			}
+
+			repo, err := f.BaseRepo()
+			if err != nil {
+				return err
+			}
+
+			dontPromptForConfirmation, err := cmd.Flags().GetBool("danger-cannot-be-undone")
```
profclems: Since we are using `--yes` flag, this should change??

commands/project/transfer/project_transfer.go
```diff
@@ -0,0 +1,91 @@
...
+
+	"github.com/MakeNowJust/heredoc"
+	"github.com/profclems/glab/commands/cmdutils"
+	"github.com/spf13/cobra"
+	"github.com/xanzy/go-gitlab"
+)
+
+func NewCmdTransfer(f *cmdutils.Factory) *cobra.Command {
+	var repoTransferCmd = &cobra.Command{
+		Use:   "transfer <command> [flags]",
```
profclems: instead of `<command>`, it should be `<repo>`.

nit: I'm also thinking we should make it optional `[repo]` so that if `repo` is not provided, we get the repo from the current git directory (with `f.BaseRepo()`)

commands/project/transfer/project_transfer.go
```diff
@@ -0,0 +1,91 @@
...
+		Short: `Transfer a repository to a new namespace.`,
+		Example: heredoc.Doc(`
+			$ glab repo transfer profclems/glab --target-namespace notprofclems
+		`),
+		Args: cobra.MaximumNArgs(2),
```
profclems: `cobra.MaximumNArgs(1)` instead??

commands/project/transfer/project_transfer.go
```diff
@@ -0,0 +1,91 @@
...
+			fmt.Fprintf(f.IO.StdOut, "%s Successfully transferred repository %s to %s\n",
+				c.GreenCheck(), c.Yellow(repo.FullName()), c.Yellow(project.PathWithNamespace))
+
+			return nil
+		},
+	}
+
+	repoTransferCmd.Flags().BoolP("yes", "y", false, "Danger: Skip confirmation prompt and force transfer operation. Transfer cannot be undone.")
+	repoTransferCmd.Flags().StringP("target-namespace", "t", "", "The namespace where your project should be transferred to")
+
```
profclems, line 89: We should make this flag compulsory since the target namespace is a requirement
```go
 	_ = repoTransferCmd.MarkFlagRequired("target-namespace")
```

profclems (changes requested):
This looks great @x3ro!

Just a few nitpicky changes, then we can merge this

profclems (approved):

x3ro:
Hey, thanks for making the last steps happen 🎉 I've been a bit short on time 😓

PR #936, opened 2022-01-08
Fix hyperlinks rendering in less

LitoMore:
## Description

We should overwrite user's configuration of less.

### Actual behavior

The two commands below outputs a incorrect string.

```shell
LESS=FRX FORCE_HYPERLINKS=1 glab mr list
```

```shell
LESS=-R FORCE=HYPERLINKS=1 glab mr list
```

### Expected behavior

GLab should overwrite user's configurations of less. We should always use `LESS=FrX` to run the less

### Actual behavior

CleanShot 2022-01-08 at 16 20 55@2x

### Expected behavior

image

## Types of changes

### DIFF
--- a/pkg/iostreams/iostreams.go
+++ b/pkg/iostreams/iostreams.go
@@ -111,12 +111,9 @@ func (s *IOStreams) StartPager() error {
 			pagerEnv = append(pagerEnv[0:i], pagerEnv[i+1:]...)
 		}
 	}
-	if _, ok := os.LookupEnv("LESS"); !ok {
-		pagerEnv = append(pagerEnv, "LESS=FRX")
-	}
-	if _, ok := os.LookupEnv("LV"); !ok {
-		pagerEnv = append(pagerEnv, "LV=-c")
-	}
+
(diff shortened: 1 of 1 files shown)


pkg/iostreams/iostreams.go
```diff
@@ -111,12 +111,8 @@ func (s *IOStreams) StartPager() error {
...
 		}
 	}
-	if _, ok := os.LookupEnv("LESS"); !ok {
-		pagerEnv = append(pagerEnv, "LESS=FRX")
-	}
-	if _, ok := os.LookupEnv("LV"); !ok {
-		pagerEnv = append(pagerEnv, "LV=-c")
-	}
+
+	pagerEnv = append(pagerEnv, "LESS=FrX", "LV=-c")
```
profclems: Maybe we should only override the user's Less config only when they set FORCE_HYPERLINK=1 ??

LitoMore: Good point.

zemzale (approved):
With the change for checking if hyperlinks should be displayed this LGTM

PR #937, opened 2022-01-08
chore: Update Arch README link to official repos

BreD1810:
## Description
`glab` has been added the the [official Arch repos](https://archlinux.org/packages/community/x86_64/glab/). This PR updates the README to reflect this.

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation
- [x] Chore (Related to CI or Packaging to platforms)

PR #938, opened 2022-01-10
chore: Correct Arch Linux install command

alerque:
* Use official package manager (pacman) instead of unsupported AUR
  helper / pacman wrapper (yay) that was only necessary for unofficial
  packaging.

* Drop db sync flag (-y) because doing that while installing packages
  results in an unsupported partial-update scenario. The correct usage
  is to update the whole system first (if necessary) with `pacman Syu`,
  but Arch users will know that, so just the single relevant install
  command is best for projects to show.


alerque:
#937 already *mentioned* the change, but this fixes the commands to match and drops the erroneous mention of "AUR helpers" since only the official package manager is necessary now.

zemzale:
Looks good

PR #951, opened 2022-02-09
Change flag which causes errors in some shells

byteknacker:
## Description

When installing glab on alpine 3.14 based Docker images, the following script throws a grep unexpected option error for the --only flag. Upon inspection of grep --help, --only does not exist. Use -o or --only-matching instead to have the same result. 

The installation command used here is:

curl -sL https://j.mp/glab-cli | sudo sh

## How Has This Been Tested?

I used this in my own Docker image based on the FROM python:3.7.12-alpine3.14 and it successfully installed version 1.22, i.e. the currently latest version of glab. 

## Types of changes

- [ ] Documentation

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -54,7 +54,7 @@ case $(uname -s) in
 esac
 
 printf "Fetching latest version\n"
-latest="$(curl -sL 'https://api.github.com/repos/profclems/glab/releases/latest' | grep 'tag_name' | grep --only 'v[0-9\.]\+' | cut -c 2-)"
+latest="$(curl -sL 'https://api.github.com/repos/profclems/glab/releases/latest' | grep 'tag_name' | grep -o 'v[0-9\.]\+' | cut -c 2-)"
 tempFolder="/tmp/glab_v${latest}"
 
 printf -- "Found version %s\n" "${latest}"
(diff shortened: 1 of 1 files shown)


scripts/install.sh
```diff
@@ -54,7 +54,7 @@ case $(uname -s) in
 esac
 
 printf "Fetching latest version\n"
-latest="$(curl -sL 'https://api.github.com/repos/profclems/glab/releases/latest' | grep 'tag_name' | grep --only 'v[0-9\.]\+' | cut -c 2-)"
+latest="$(curl -sL 'https://api.github.com/repos/profclems/glab/releases/latest' | grep 'tag_name' | grep -o 'v[0-9\.]\+' | cut -c 2-)"
```
zemzale: Could maybe change this to the `--only-matching`, since in scripts it's better to have the full names, since it' s easier to read and understand it

zemzale (changes requested):
Thanks for the PR. I have only one small change I would like to see.

byteknacker:
> Thanks for the PR. I have only one small change I would like to see.

Just made the change to the verbose flag. Does that look good?

zemzale (approved):
LGTM

byteknacker:
Ok, I just found out that for me at least, this --only-matching causes errors. The grep in the base docker image I am using does not recognise this flag, only -o is recognised. Cannot show details since it is for my client.  I just want to point out that for some -o is better than --only-matching.

Issue #872, opened 2021-10-12
Support for snippets


darix:
creating/listing/deleting would be nice

and for creating something like `somecommand | glab snippet new -` would be nice.

clemsbot:
This issue has been automatically marked as stale because it has not had recent activity. We haven't had the time to address it yet, but we want to keep it open. This message is just a reminder for us to help triage issues.

georgexsh:
+1 for this

zemzale:
I am gonna slowly started working on this. Won't be very fast since I don't have that much time to work on this, but I hope for some slow and steady progress

Lucas-C:
I'd love to have this feature too!

Given that [Drag and Drop](https://gitlab.com/gitlab-org/gitlab/-/issues/223940) is not implemented for now on Gitlab side, I'd be great to be able to `glab snippet new file1.py file2.rb` in order to quickly create a snippet ! And to update it: `glab snippet updatefile1.py file2.rb` maybe?

clemsbot:
This issue has been automatically marked as stale because it has not had recent activity. We haven't had the time to address it yet, but we want to keep it open. This message is just a reminder for us to help triage issues.

zemzale:
So I have created the initial support for adding snippets, but that only works for a single file, because the GitLab API library didn't support it when writing the initial PR, but it does now since I added that there. So there are a few things to do
- [ ] Update the GitLab API library to the newest version to support multiple file uploads
- [ ] Add the support for multiple files in the create command
- [ ] Add missing commands 
  - [ ] Update command
  - [ ] View command
  - [ ] Delete command

Lucas-C:
Wow, great, thanks for working on this @zemzale !

clemsbot:
This issue has been automatically marked as stale because it has not had recent activity. We haven't had the time to address it yet, but we want to keep it open. This message is just a reminder for us to help triage issues.

PR #952, opened 2022-02-10
feat(snippet): Add command for creating snippets

zemzale:
## Description
This adds a simple snippet creation command. This currently just allows
for creating snippets from STDIN like so:

`echo "this is a script" | glab snippet create --title "snippet" --filename "example.py"`

There are few things that are missing from this, like ability to add
more then one file to the snippet, but that is not supported by the "github.com/xanzy/go-gitlab" yet.

Issue is opened for that https://github.com/xanzy/go-gitlab/issues/1372

There are gonna be follow up MRs to add some extra things like:
- creating personal snippets
- creating snippets with multiple files
- adding snippets from files instead of STDIN

Even with those features missing this should be a good start

## Related Issue

Issue #872

## How Has This Been Tested?

Local tests and some unit tests

## Types of changes

- [ ] Documentation

### DIFF
--- a/commands/snippet/create/create.go
+++ b/commands/snippet/create/create.go
@@ -1,0 +1,145 @@
+package create
+
+import (
+	"errors"
+	"fmt"
+	"io"
+	"os"
+
+	"github.com/MakeNowJust/heredoc"
+	"github.com/profclems/glab/api"
+	"github.com/profclems/glab/commands/cmdutils"
+	"github.com/profclems/glab/internal/glrepo"
+	"github.com/profclems/glab/pkg/iostreams"
+	"github.com/spf13/cobra"
+	"github.com/xanzy/go-gitlab"
+)
+
+type CreateOpts struct {
+	Title           string
+	Description     string
+	DisplayFilename string
+	Visibility      string
+
+	ForUser  bool
+	FilePath string
+
+	IO       *iostreams.IOStreams
+	Lab      func() (*gitlab.Client, error)
+	BaseRepo func() (glrepo.Interface, error)
+}
+
+func (opts CreateOpts) isSnippetFromFile() bool {
+	return opts.FilePath != ""
+}
+
+func NewCmdCreate(f *cmdutils.Factory) *cobra.Command {
+	opts := &CreateOpts{}
+	var snippetCreateCmd = &cobra.Command{
+		Use:     "create [path]",
+		Short:   `Create new snippet`,
+		Long:    ``,
+		Aliases: []string{"new"},
+		Example: heredoc.Doc(`
+			$ glab snippet create script.py --title "Title of the snippet" 
+			$ echo "package main" | glab snippet new --title "Title of the snippet" --filename "main.go"
+			$ glab snippet create main.go -t Title -f "different.go" -d Description
+		`),
+		PreRunE: func(cmd *cobra.Command, args []string) error {
+			opts.IO = f.IO
+			opts.BaseRepo = f.BaseRepo
+			opts.Lab = f.HttpClient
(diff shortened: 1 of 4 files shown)


commands/snippet/create/create.go
```diff
@@ -0,0 +1,120 @@
...
+			$ glab snippet create -t Title -f "main.go" -d Description
+		`),
+		Args: cobra.ExactArgs(0),
+		PreRunE: func(cmd *cobra.Command, args []string) error {
+			opts.IO = f.IO
+			opts.BaseRepo = f.BaseRepo
+			opts.Lab = f.HttpClient
+			if opts.Title == "" {
+				return &cmdutils.FlagError{
+					Err: errors.New("--title required for snipets"),
```
profclems: nit: typo "snippets"

commands/snippet/create/create.go
```diff
@@ -0,0 +1,120 @@
...
+				}
+			}
+			if opts.Filename == "" {
+				return &cmdutils.FlagError{
+					Err: errors.New("--filename required for snipets"),
```
profclems: nit: typo "snippets"

commands/snippet/create/create.go
```diff
@@ -0,0 +1,120 @@
...
+
+	IO       *iostreams.IOStreams
+	Lab      func() (*gitlab.Client, error)
+	BaseRepo func() (glrepo.Interface, error)
+}
+
+func NewCmdCreate(f *cmdutils.Factory) *cobra.Command {
+	opts := &CreateOpts{}
+	var snippetCreateCmd = &cobra.Command{
+		Use:     "create",
```
profclems: We should allow an optional file path to be specified as an argument `create [path]`. When no path is specified as argument, we fall back to stdin. `--filename` should be optional when `path` is specified so we should the filename as default.

This will help us when the API finally supports multiple snippet files.

profclems (changes requested):
This is great @zemzale! I'm so excited about this. Just a few nit comments to get this merged

commands/snippet/create/create.go
zemzale: You right. Gonna try to get these changes in

zemzale: So I changed the logic a bit. 

Now we can take in an argument of the path. In these cases, the `filename` flag is not required, but still can be used to override the display name in GitLab. 

I split the file reading functions into easily readable parts that make them easier to read. 
Also added a simple function that can tell me if the snippet is coming from a file or STDIN. I think this makes the intent of the code a lot clearer.

commands/snippet/create/create.go
zemzale: Fixed this

commands/snippet/create/create.go
zemzale: Fixed this

commands/snippet/snippet_test.go
```diff
@@ -0,0 +1,41 @@
...
+	"github.com/alecthomas/assert"
+	"github.com/profclems/glab/commands/cmdtest"
+	"github.com/profclems/glab/commands/cmdutils"
+)
+
+func TestMain(m *testing.M) {
+	cmdtest.InitTest(m, "mr_cmd_test")
+	cmdtest.InitTest(m, "mr_cmd_autofill")
+}
+
```
profclems: nit: you can remove this

profclems (approved):

commands/snippet/snippet_test.go
zemzale: Copy-pasting at its best. Removed it.

PR #971, opened 2022-03-08
fix: fix validation for GitLab 14.5 tokens

### LINKED ISSUE #926: invalid token provided
**Description**

I have generated a new gitlab personal access token 

I used glab auth login to authenticate

```sh
➜ glab auth login 
? What GitLab instance do you want to log into? gitlab.com
- Logging into gitlab.com
? Paste your authentication token: **************************
? Choose default git protocol SSH
- glab config set -h gitlab.com git_protocol ssh
✓ Configured git protocol
- glab config set -h gitlab.com api_protocol https
✓ Configured API protocol
✓ Logged in as ************
```

auth status check reports that i have provided an invalid token

```sh
 ➜ glab auth status
gitlab.com
  ✓ Logged in to gitlab.com as ************ (/Users/************/.config/glab-cli/config.yml)
  ✓ Git operations for gitlab.com configured to use ssh protocol.
  ✓ API calls for gitlab.com are made over https protocol
  ✓ REST API Endpoint: https://gitlab.com/api/v4/
  ✓ GraphQL Endpoint: https://gitlab.com/api/graphql/
  ✓ Token: ********************
  ! Invalid token provided
```

while in this state issuing glab commands on a repo returns the message

none of the git remotes configured for this repository points to a known GitLab host. Please use `glab auth login` to authenticate and configure a new host for glab

glab should report the token as valid

**Possible Fix**
is it possible that token format has recently changed adding glpat- prefix

**Steps to Reproduce**
provided in description

**Logs**
provided in description

**Your Environment**
* Version used (Run `glab --version`):

glab version 1.21.1 (2021-09-28)

MacOS 10.6.1

ygarasab:
## Description

- update token validation both in token length and prefix (now standard on gitlab)

- update documentation for token validation

- update '*' based token alias showed in log for proper length

This solves the current gitlab.com issue with the command

`glab auth status`

But there are still some warning messages related to some db.org  and gl.io domains, but I don't really know what that's about :(

## Related Issue

Resolves #926

## How Has This Been Tested?

Local testing

## Types of changes

- [ ] Documentation

### DIFF
--- a/api/default.go
+++ b/api/default.go
@@ -3,10 +3,10 @@ package api
 var DefaultListLimit = 30
 
 //IsValidToken checks if a token provided is valid.
-// The token string must be 20 characters in length to be recognized as
-// a valid personal access token.
+// The token string must be 26 characters in length and have the 'glpat-'
+// prefix to be recognized as a valid personal access token.
 //
 // TODO: check if token has minimum scopes required by glab
 func IsValidToken(token string) bool {
-	return len(token) == 20
+    return len(token) == 26 && token[:6] == "glpat-"
 }
--- a/api/default.go
+++ b/api/default.go
@@ -4,9 +4,9 @@ var DefaultListLimit = 30
 
 //IsValidToken checks if a token provided is valid.
 // The token string must be 26 characters in length and have the 'glpat-'
-// prefix to be recognized as a valid personal access token.
+// prefix or just be 20 characters long to be recognized as a valid personal access token.
 //
 // TODO: check if token has minimum scopes required by glab
 func IsValidToken(token string) bool {
-    return len(token) == 26 && token[:6] == "glpat-"
+    return (len(token) == 26 && token[:6] == "glpat-") || len(token) == 20
 }
--- a/api/default.go
+++ b/api/default.go
@@ -8,5 +8,5 @@ var DefaultListLimit = 30
 //
 // TODO: check if token has minimum scopes required by glab
 func IsValidToken(token string) bool {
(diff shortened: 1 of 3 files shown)


api/default.go
```diff
@@ -3,10 +3,10 @@ package api
...
 //IsValidToken checks if a token provided is valid.
-// The token string must be 20 characters in length to be recognized as
-// a valid personal access token.
+// The token string must be 26 characters in length and have the 'glpat-'
+// prefix to be recognized as a valid personal access token.
 //
 // TODO: check if token has minimum scopes required by glab
 func IsValidToken(token string) bool {
-	return len(token) == 20
+    return len(token) == 26 && token[:6] == "glpat-"
```
profclems: This would break older versions of self-hosted GitLab instances. 

We could still validate both newer and older versions:
```go
(len(token) == 26 && token[:6] == "glpat-") || len(token) == 20
```

profclems: Also make sure to run gofmt on the change or you can run
`make fix`

profclems (changes requested):
@ygarasab thanks so much for working on this fix! I’m excited about this

api/default.go
ygarasab: ok, i'll check it out, thx for the patience

I was looking at the tests right now and I was wondering if I should add one with the new token standard, guess I'll do that now

ygarasab:
I'm gonna give a go at the tests :D, never tested in go though

zemzale (approved):
LGTM! Thanks for the patch @ygarasab

Issue #974, opened 2022-03-10
Creating a merge request: assignee, reviewer and label do not work


lpapp:
Hi,

I am on Mac OS (Monterey 12.2.1.).

I installed glab through brew.

`glab auth login` works fine.

When I am trying to create a merge request, what happens is that -a laszlo.papp creates /assignee laszlo.papp in the description (body) of the merge request. Why is this happening as opposed to selecting me as the assignee in the Assignee drop-down list?

We are using this gitlab version on the server: https://gitlab.com/gitlab-org/gitlab/-/tags/v14.0.12-ee

```
glab --version
glab version 1.22.0 (2022-01-10)
```

This is the command I am trying to use:

`glab mr create -a laszlo.papp -b target_branch -f -w`

The username is correct because if I try to remove one character from it, I am getting an error. Also, this is the user name used for authentication at login (git auth login) which also works.

I wanted to get more information, but I cannot see a --verbose, --trace, --debug or some similar option.

Would you be able to help me out with this? Thanks.

zemzale:
Hi @lpapp 

What is happening here is that when using the `-w` flag you will get redirected to a web browser where you can continue with your MR creation. If we continue to the browser the only way how we can add the assigne is using the slash commands, so we just append them to the MR description. 

See https://docs.gitlab.com/ee/user/project/quick_actions.html

The reason for this is that there is no API that allows to create "preview" MR, so we have to add all the metadata you provided using these quick actions.

lpapp:
@zemzale oh, I see, thanks! So, the issue is that there is no preview API provided by the gitlab server.

So, then, the only issue that I can see is that this option on the command line:

`--reviewer foo.bar`

does not get added on the preview as:

`/reviewer foo.bar`

like the `-a foo.bar` options gets added as:

`/assign foo.bar`

or the `-l "foo" `as:

`/label ~"foo"`

Does it work for you? Am I doing something wrong?

zemzale:
The reviewer is missing indeed. All the other things should be correct. 

If you press create MR on the Web UI it should add all the labels and assignes. I will try to open a PR to fix the reviewer thing.

lpapp:
I can open a PR for this. It would be my first go contribution in a long time :) I think it is as simple as this:

git diff
diff --git a/commands/mr/create/mr_create.go b/commands/mr/create/mr_create.go
index c66ba99..4cbe44d 100644
--- a/commands/mr/create/mr_create.go
[...]

PR #975, opened 2022-03-10
feat: provide /reviewer quick action

lpapp:
This is when creating a merge request using the command line option
git mr create --reviewer -w. Other options, like -a and -l, have been
working, but this seems to have gone missed.

Delivers #974


zemzale (approved):
Thanks for the fix @lpapp. LGTM

PR #976, opened 2022-03-10
docs: change --labels to --label

lpapp:
The correct parameter seems to be --label (singular) as opposed to
--labels (plural).

It is a bit strange that the same example is given in two different
places, so this may need centralisation in one place in the future.


docs/source/mr/index.rst
```diff
@@ -16,7 +16,7 @@ Examples
 
 ::
 
-  $ glab mr create --fill --labels bugfix
+  $ glab mr create --fill --label bugfix
```
zemzale: You don't need to update this. It's automatically generated. The change in the Example is enough. So if u can remove this change this is good to go

zemzale (changes requested):

zemzale (approved):

PR #982, opened 2022-03-17
chore update go gitlab

### LINKED ISSUE #872: Support for snippets

zemzale:
## Description

This PR updates go-gitlab library to the newest version so we can add more then
one file to snippets. 

The problem here is that there is one huge nasty change that changes a lot of slices to be pointers to slices [see here](https://github.com/xanzy/go-gitlab/pull/1294)

There are also some good things as, no need to use the workaround for setting
the scope for environment filters when removing project variables. 

There is also a change in the way how labels are treated in most places (this
actually just looks like a normalization around the library since some places
already used this custom type and some didn't).

There is quite a lot of pointer derefrencing now, and I have not done any real
testing so there might be some problems so this probably should be tested better
before merging.

Each change should be in it's own commit for easier review, but should be
squashed before merge, because most of the commits fail to build and I don't
like having commits with build failures. 

Commit messages should have better explanation for most changes.

## Related Issue

Resolves #872

## Types of changes

- [ ] Documentation


ygarasab:
hey, I've been wanting to ask this for a while, but what is the best way to set the local environment for running the tests properly?

